PR #212, opened 2015-03-06
Include '<netinet/in.h>' for using 'struct sockaddr_in'

syohex:
This is related to #211.


dch:
+1 per my comment on #211.

PR #405, opened 2015-07-19
Add fail handler

fbbdev:
Fail handler sends an error response with the specified status. This is useful to prevent clients from accessing non-existent hosts or protected paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 CMakeCache.txt
 CMakeFiles/
 Makefile
+*.gch
 build/
 cmake_install.cmake
 xcuserdata
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,6 +191,7 @@ SET(LIB_SOURCE_FILES
     lib/handler/access_log.c
     lib/handler/chunked.c
     lib/handler/expires.c
+    lib/handler/fail.c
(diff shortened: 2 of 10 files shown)


fbbdev:
# Examples

## Prevent access to non-existent hosts
```yaml
hosts:
  default:
    paths:
      /:
        fail: 403
  "example.com:80":
    paths:
      /:
        file.dir: /path/to/doc-root
```

## Hide details
```yaml
hosts:
  default:
    paths:
      /:
        file.dir: /path/to/doc-root
        file.index: [ 'foo.html' ]
      /foo.html:
        fail: 404
```

kazuho:
Thank you for the PR!  And thank you for including documentation and tests.

Please let me consider how I should merge this.

kazuho:
Thank you for waiting long.

To keep the core small (and to provide maximum flexibility), I have decided not to merge this PR but instead to encourage people to use scripting with mruby for this kind of requirements.

Using the mruby handler, you can write like:

```
paths:
  /:
    mruby.handler: |
      H2O.return 403, "Access Forbidden", "I will never show you this!"
```

And once #471 gets merged, it would become possible to send a file with an arbitrary status code (using mruby).

fbbdev:
Thank you.

PR #830, opened 2016-03-04
Enhance description of headers about http2 compatibility

mingodad:


mingodad:
It doesn't make sense the server should manage this internally

PR #1027, opened 2016-08-15
use examp1e.net in tests instead of xip.io

i110:
Tests often failes due to domain name resolution problem, so we should use `examp1e.net`, which we can control, instead of `xip.io`

As I told, please setup so that `alternate.127.0.01.examp1e.net` is resolved to `127.0.0.1` (like xip.io)


i110:
@kazuho As I told, it seems to be better to use `Custom Host Names` feature of Travis CI, so close this PR and create another PR.
https://docs.travis-ci.com/user/hosts

PR #1143, opened 2016-12-20
add session_ticket

9il:
Still have some problems with linking new OpenSSL.

LibreSSL has not SSL_SESSION_get0_ticket. But if we need it, I will look if tickets can be accessed directly.

Do we need session_id?

The following info is for client code and we can not use it in server. SSL object or SSL_SESSION should  be used instead.
```c
            struct {
                char *server_name;
                h2o_cache_t *session_cache;
                h2o_iovec_t session_cache_key;
                h2o_cache_hashcode_t session_cache_key_hash;
            } client;
```


9il:
wrong repo

PR #1277, opened 2017-04-24
Support load balancing

zlm2012:
To support load balancing, I think we should do the following jobs:

- [x] support load balancing in lib/common/socketpool.c
- [x] modify directive for proxy so we can support multiple remote destinations & modify handler for proxy so we can use the modified socket pool to achieve load balancing.

Currently, socket pool should barely support multiple destinations structurally by this PR. Still a lot of things to do.


zlm2012:
Last failed build is caused by "address is already in use", which should be a race condition.

include/h2o/socketpool.h
```diff
@@ -67,11 +81,20 @@ typedef struct st_h2o_socketpool_t {
         pthread_mutex_t mutex;
         h2o_linklist_t sockets; /* guarded by the mutex; list of struct pool_entry_t defined in socket/pool.c */
     } _shared;
+
+    /* vars used by load balancing, modified by multiple threads */
+    struct {
+        h2o_socketpool_lb_selector selector;
+        h2o_socketpool_lb_dispose_cb dispose;
+        pthread_mutex_t mutex;
```
deweerdt: since the mutex might or might not be needed by the load balancing policy, it might make sense to move it to `struct round_robin_t` and have it locked in `lb_selector`?

zlm2012:
Mutex has been moved within balancer, which is a much more natural way. Though build still failed which looks like a race condition again? I'm currently not sure what happened. I've run CI build again on my own branch & passed.

deweerdt:
> Though build still failed which looks like a race condition again?

deweerdt:
> Mutex has been moved within balancer, which is a much more natural way. 

Thank you for the changes! 92919cf looks good to me.

deweerdt:
@zlm2012 The changes look good to me, there's a conflict in `lib/handler/mruby/http_request.c`, do you mind taking a look? The title still includes `[WIP]`, is that still true?

zlm2012:
Conflict has been resolved. Then we need to wait for the Travis CI result. `[WIP]` was forgotten to be removed.

deweerdt:
Thank you for your efforts, this is a fantastic addition to H2O.

PR #1296, opened 2017-05-12
Remove superfluous san/cov flags for fuzz builds

jfoote:
The existing `CMakeLists.txt` contains a superfluous linker directive for `-fsanitize=address`. google/oss-fuzz has recently started experimenting with additional Clang sanitizers; [this directive causes build errors](https://github.com/google/oss-fuzz/issues/585) when used with other sanitizer/coverage options.

This patch removes the superfluous directive and fixes an error in `fuzz/driver.cc` that was causing memsan fuzz runs to report a bug and exit immediately.


deweerdt:
lgtm, thanks!

inferno-chromium:
Thanks!

PR #1312, opened 2017-05-25
Add flags for linker

jwinarske:
When building h2o as an external project via git submodule, there is a need to override 

    FIND_PACKAGE(OpenSSL)

In this case there are some scenarios where the linker needs "-lssl -crypto".  This appears to be isolated to OpenSSL.

This is an example use case of overriding the auto package find behavior:

    set(h2o_src_path "${CMAKE_SOURCE_DIR}/h2o")
    set( LIBH2O_INCLUDE_DIR "${h2o_src_path}/include")
    set( LIBH2O_LIBRARY     "${h2o_src_path}/libh2o.a")
    ExternalProject_Add(h2o
          SOURCE_DIR ${h2o_src_path}
          DOWNLOAD_COMMAND ""
          BUILD_IN_SOURCE 1
          CMAKE_ARGS -DWITH_MRUBY=OFF -DWSLAY_FOUND=1 -DWSLAY_INCLUDE_DIRS=${WSLAY_INCLUDE_DIR} -DWSLAY_LIBRARIES=${WSLAY_LIBRARIES} -DOPENSSL_FOUND=1 -DOPENSSL_INCLUDE_DIRS=${OPENSSL_INCLUDE_DIR} -DOPENSSL_CRYPTO_LIBRARY=${OPENSSL_CRYPTO_LIBRARY} -DOPENSSL_SSL_LIBRARY=${OPENSSL_SSL_LIBRARY} -DOPENSSL_LIBRARIES=${OPENSSL_LIBRARIES} -DLIBUV_FOUND=1 -DLIBUV_VERSION=1.11.0 -DLIBUV_INCLUDE_DIR=${LIBUV_INCLUDE_DIR} -DLIBUV_LIBRARIES=${LIBUV_LIBRARIES}
          INSTALL_COMMAND ""
    )


INRIX-joel-winarske:
I found out that the -lssl and -lcrypto flags pull in the system openssl beyond what CMake refers to!  This is bad, as the linker will mark your binary "needing" two different OpenSSL instances.  So if you override with version X that is different what the system has, it will have two different versions "needed".  Regardless this is bad and I'm removing this PR.

PR #1362, opened 2017-07-09
add modern default ciphers to example configs

yannick:
since https://github.com/h2o/h2o/pull/1334 hasn't found an agreement we should at least encourage people to take modern defaults.


yannick:
Q: should we add oscp and session ticket settings?
pro: raise awareness
contra: clutter 

another option would be to provide a manually generated full fledged config file with all defaults + comments being set.
currently it's extremely hard to find those on a machine sans html reader:
```
% man h2o
No manual entry for h2o
```
(another reason why it would be  nice to be able to view a full running config)

PR #1400, opened 2017-08-17
Forward Date header to upstream server

ykzts:
I'm a committer of Mastodon (Rails web application).

Mastodon has the ability to connect with remote instances using [ActivityPub](https://www.w3.org/TR/activitypub/). ActivityPub validates connection using [Signing HTTP Message](https://w3c-dvcg.github.io/http-signatures/).

When using H2O as a reverse proxy, the Date header does not reach the upstream server. Therefore, verification of connection always fails.

Forward the Date header contained in the request header to the upstream server.


deweerdt:
:+1: aff6a96

ykzts:
I think #1408 is better.

i110:
@ykzts Thanks for filing this PR. But I think that this may causes additional issues around response header dropping.
I created https://github.com/h2o/h2o/pull/1408, and it addresses these issues

i110:
@ykzts #1408 gets merged, so please checkout and use the latest master to resolve your issue.

ykzts:
@i110 @kazuho Thanks!

PR #1442, opened 2017-10-04
fix test issues

i110:
- `t/50reverse-proxy-chunk-timeout-2.t` uses nghttp without precondition check
- in `t/50reverse-proxy-serialize-posts.t`:
  - grep command in OSX no longer supports `P` option
  - `wc` command possibly outputs leading white spaces

This PR fixes above issues


deweerdt:
LGTM, thank you for the fixes!

PR #1484, opened 2017-10-27
WIP document mruby redis feature

yannick:
its a bit opinionated example that shows other things, but i think since grasping the redis interface is so simple it is ok


doc/configure/mruby.html
```diff
@@ -269,6 +269,57 @@ <h3 id="logging-arbitrary-variable">Logging Arbitrary Variable</h3>
...
+            user_name, threads = [
+              redis.get(&quot;users:#{user_id}&quot;),
+              redis.get(&quot;numberofthreads&quot;),
+            ].map(&amp;:join) #parallel calls
+            result_string = {&#39;name&#39; =&gt; user_name, &quot;server_threads&quot; =&gt; threads }.to_json
+            [200, {&quot;Content-Type&quot; =&gt; &quot;application/json&quot;}, [result_string]]
+          end
+          #curl &#39;localhost:8080/test?user_id=1&#39;
+</code></pre>
+</div>
```
i110: One thing that might be worth noting is connection management.

1. If the user wants to prepare connection beforehand (typically at startup), he can call `connect` method explicitly, otherwise the connection will be established immediately before the first command is sent
2. If the user wants to disconnect connection per request, he can call `redis.disconnect` (it's often needed if he is suffering from `Too many connection` error)

Taking the above into account, how about adding the following example with a concise explanation?
```ruby
mruby.handler: |
  redis = H2O::Redis.new(:host => '127.0.0.1', :port => 6379)
  redis.connect
  proc {|env|
    reply = redis.get('foo').join
    redis.disconnect # if you want to disconnect per request
    [200, {}, [reply]]
  }
```

yannick:
option would be to remove the pid stuff from example 2, but it makes it less fun.

PR #1516, opened 2017-12-03
# Changed the window size update on data frame to update both the str…

objd:
…eam's window size and the connection's window size.

This fixes an issue where big files could not be uploaded, as bad flow control values prevented the client from sending more data frames.

### DIFF
--- a/lib/http2/connection.c
+++ b/lib/http2/connection.c
@@ -630,8 +630,6 @@ static int handle_data_frame(h2o_http2_conn_t *conn, h2o_http2_frame_t *frame, c
     if (conn->state >= H2O_HTTP2_CONN_STATE_HALF_CLOSED)
         return 0;
 
-    update_input_window(conn, 0, &conn->_input_window, payload.length);
-
     /* save the input in the request body buffer, or send error (and close the stream) */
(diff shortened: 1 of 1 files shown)


kazuho:
@sturcotte06 @objd Thank you for the PR.

Could it be the case that this is an issue in the handler you are writing?

When using streaming request mode, it is the handler's responsibility to call the `proceed_req` callback as it consumes the stream. And that is the only moment when the HTTP/2 protocol handler updates the stream-level flow control window (see https://github.com/h2o/h2o/blob/master/lib/http2/connection.c#L612).
Otherwise, H2O will be required to buffer non-neglible amount of data if client sends data faster than the handler consumes.

You might be interested in looking the reverse proxy implementation (lib/core/proxy.c) that uses the mode.

sturcotte06:
Yes it could very well be my implementation that is wrong. So I should be
calling the proceed_req callback on every chunk received?

sturcotte06:
Actually, I've tried all weekend to diagnose was wrong. I've understood it
at the protocol level, because I could see the http2 frames coming, but I
was unable to find any examples of a simple chunked handler. I'll try to
call proceed req in my handler; if that works, I'll close the issue and the
PR.

Great project by the way =)

PR #1549, opened 2017-12-16
[WIP] http2client

i110:
TODOs:

- [ ] request body streaming
- [ ] automatic retry (when received goaway or REFUSED_STREAM)
- [ ] decode `transfer-encoding: chunked` response
- [ ] fix connection selecting logic (least-streams with certain round-robin manner?)
- [ ] merge with http1client into `httpclient`
- [ ] write tests


i110:
@kazuho Could you review this when you have a time?

include/h2o/httpclient.h
```diff
@@ -0,0 +1,112 @@
...
+
+struct st_h2o_header_t;
+
+typedef void (*h2o_httpclient_proceed_req_cb)(h2o_httpclient_t *client, size_t written, int is_end_stream);
+typedef int (*h2o_httpclient_body_cb)(h2o_httpclient_t *client, const char *errstr);
+typedef h2o_httpclient_body_cb (*h2o_httpclient_head_cb)(h2o_httpclient_t *client, const char *errstr, int minor_version,
+                                                           int status, h2o_iovec_t msg, struct st_h2o_header_t *headers,
+                                                           size_t num_headers, int rlen);
+typedef h2o_httpclient_head_cb (*h2o_httpclient_connect_cb)(h2o_httpclient_t *client, const char *errstr, h2o_iovec_t *method, h2o_url_t *url,
+                                                              h2o_headers_t *headers, h2o_iovec_t *body,
```
kazuho: Considering the fact that we will be passing back a _constant_ list of headers, the argument should be `const h2o_header_t **headers, size_t *num_headers` instead of `h2o_headers_t *headers`.

examples/libh2o/http1client.c
```diff
@@ -39,64 +39,59 @@ static h2o_iovec_t iov_filler;
...
+                                         h2o_httpclient_features_t features, h2o_url_t *origin);
+static h2o_httpclient_body_cb on_head(h2o_httpclient_t *client, const char *errstr, int minor_version, int status, h2o_iovec_t msg,
+                                      h2o_header_t *headers, size_t num_headers, int rlen);
 
-static void start_request(h2o_http1client_ctx_t *ctx)
+static void start_request(h2o_httpclient_ctx_t *ctx)
 {
-    h2o_url_t url_parsed;
-    h2o_iovec_t *req;
+    h2o_url_t *url_parsed;
```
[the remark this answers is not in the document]
i110: > Is it due to an API change from http1client to httpclient, or is it due to an existing bug becoming apparent?

Since I wanted to store `url` in `client->data` for later use. Original code stored HTTP/1.1 request buffer, so this can be said as the change caused by API changes. Anyway this is an example, so I can change this easily if you like.

include/h2o/httpclient_internal.h
```diff
@@ -0,0 +1,53 @@
...
+ * IN THE SOFTWARE.
+ */
+#ifndef h2o__httpclient_internal_h
+#define h2o__httpclient_internal_h
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "h2o.h"
```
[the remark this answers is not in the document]
i110: This is only for sharing the definitions of internal structures between `http*client.c`. So should I move this file to elsewhere e.g. under `lib/common/` ?

examples/libh2o/http1client.c
```diff
@@ -158,44 +153,48 @@ static void timeout_cb(h2o_timeout_entry_t *entry)
...
-    *reqbufs = (h2o_iovec_t *)client->data;
-    *reqbufcnt = 1;
-    *method_is_head = 0;
+    *_method = h2o_iovec_init(method, strlen(method));
+    *url = *((h2o_url_t *)client->data);
+
     if (cur_body_size > 0) {
+        char *clbuf = h2o_mem_alloc_pool(&pool, char, sizeof(H2O_UINT64_LONGEST_STR) - 1);
+        size_t clbuf_len = sprintf(clbuf, "%d", cur_body_size);
+        h2o_add_header(&pool, headers, H2O_TOKEN_CONTENT_LENGTH, NULL, clbuf, clbuf_len);
```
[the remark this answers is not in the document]
i110: I think this is the discussion of usability vs. controllability. I felt either's fine, but it'll affect mruby's http_request, so I kept the original http1client.c design (i.e. preferred controllability). What do you think?

include/h2o/socketpool.h
```diff
@@ -176,7 +180,7 @@ void h2o_socketpool_unregister_loop(h2o_socketpool_t *pool, h2o_loop_t *loop);
  * connects to the peer (or returns a pooled connection)
  */
 void h2o_socketpool_connect(h2o_socketpool_connect_request_t **_req, h2o_socketpool_t *pool, h2o_url_t *url, h2o_loop_t *loop,
-                            h2o_multithread_receiver_t *getaddr_receiver, h2o_socketpool_connect_cb cb, void *data);
+                            h2o_multithread_receiver_t *getaddr_receiver, h2o_iovec_t alpn_protos, h2o_socketpool_connect_cb cb, void *data);
```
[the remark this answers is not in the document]
i110: I think we have to select h1 or h2 per request due to load balancing between those. Do you mean we 

> because a socketpool is expected to contain a set of sockets that are used for certain protocol.

If I recall collectly, in our previous discussion we agreed that we use one socketpool for both getting h1 and h2 sockets, doesn't it?

lib/common/http1client.c
```diff
@@ -80,13 +45,15 @@ static void close_client(struct st_h2o_http1client_private_t *client)
...
     if (client->_body_buf_in_flight != NULL)
         h2o_buffer_dispose(&client->_body_buf_in_flight);
-    free(client);
+    h2o_mem_clear_pool(&client->pool);
+    free(common);
 }
 
 static void on_body_error(struct st_h2o_http1client_private_t *client, const char *errstr)
 {
+    struct st_h2o_httpclient_private_t *common = H2O_STRUCT_FROM_MEMBER(struct st_h2o_httpclient_private_t, http1, client);
```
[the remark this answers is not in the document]
i110: Because there're two types of *client* here: `st_h2o_http1client_private_t` and `st_h2o_httpclient_private_t`. Calling both `client`s got me too confused.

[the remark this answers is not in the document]
kazuho: What I would do here is rename `struct st_h2o_http1client_private_t *client` to `_client`, refer to `struct st_h2o_httpclient_private_t *` as `client`, and never use `_client` once we have `client`.

This is because the members of `struct st_h2o_http1client_private_t` can always be referred to as `client->super.*`.

Using two variables that have a delta of some constant is an unnecessary complication, both at the source code-level and in machine code.

include/h2o/httpclient_internal.h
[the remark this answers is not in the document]
kazuho: FWIW there is a file named `include/h2o/http2_internal.h`. Maybe we can follow the convention?

examples/libh2o/http1client.c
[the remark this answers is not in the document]
kazuho: Thank you for the clarification.

> Since I wanted to store `url` in `client->data` for later use.

I assume that you are passing `url` which is a pointer to `h2o_url_t *` as an argument to `h2o_httpclient_connect`. Assuming that that is true, I do not like this change, because it sounds like a silent API change where the lifetime of the argument that the caller needs to guarantee changes from "while the function is called" to sometime after than that. Generally speaking that is a dangerous change, because it is so easy to miss fixing existing code that relies on the current behavior.

lib/common/http1client.c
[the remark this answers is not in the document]
i110: > This is because the members of struct st_h2o_http1client_private_t can always be referred to as client->super.*.

No, st_h2o_http1client_private_t can be referred as `client->http1`. One reason I introduced this variable is for consistency with `http2client.c: Please see https://github.com/h2o/h2o/blob/a697525fca5b887f92e08e508aa35da0baf31e5e/lib/common/http2client.c#L124. So if we change as you said, should I rewrite http2client.c too not to refer `stream`, and use `client->http2` style everywhere?

include/h2o/socketpool.h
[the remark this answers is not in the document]
kazuho: > If I recall collectly, in our previous discussion we agreed that we use one socketpool for both getting h1 and h2 sockets

Ah. You are correct. Thank you for pointing that out.

Then, could we have a doc-comment describing how the parameter should be used? I'd assume that the proper behavior when passing multiple protocol names is that you should only return connections that share "one" protocol name (e.g. h1) and close the connections that uses any other negotiated protocol.

examples/libh2o/http1client.c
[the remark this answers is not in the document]
i110: NOTE: As we discussed, This is not an API change, but the change caused by API changes. So this isn't dangerous and we left this as it is.

misc/tokens.pl
```diff
@@ -77,6 +77,36 @@
...
+
+/**
+ * returns a token (an optimized subclass of h2o_iovec_t) containing given string, or NULL if no such thing is available
+ */
+const h2o_token_t *h2o_lookup_token(const char *name, size_t len);
+/**
+ * returns an boolean value if given buffer is a h2o_token_t.
+ */
+int h2o_iovec_is_token(const h2o_iovec_t *buf);
+
```
kazuho: I feel uneasy about having definitions in an automatically generated file, because it makes us hard to change these definitions.

How about doing the following:
* change the name of the header file being generated from `include/h2o/token.h` to `include/h2o/token_table.h`
* define these functions and types in `h2o/token.h`, as well as including `token_table.h`

include/h2o/httpclient.h
```diff
@@ -0,0 +1,109 @@
...
+#endif
+
+#include "h2o/header.h"
+#include "h2o/socket.h"
+#include "h2o/socketpool.h"
+#include "h2o/timeout.h"
+
+typedef struct st_h2o_httpclient_t h2o_httpclient_t;
+
+typedef struct st_h2o_httpclient_features_t {
```
kazuho: What is the role of the struct?

It is confusing to me, because if it represents a feature set of a HTTP client implementation, then I do not see why it needs to be pointer.

include/h2o/httpclient.h
```diff
@@ -0,0 +1,109 @@
...
+    h2o_iovec_t *proxy_protocol;
+    int *chunked;
+    int connection_header;
```
kazuho: I can anticipate that `chunked` is a boolean. But I cannot understand what the type or the role of `connection_header` is. I'd appreciate it if you could add comments or change the name.

include/h2o/httpclient.h
```diff
@@ -0,0 +1,109 @@
...
+typedef h2o_httpclient_body_cb (*h2o_httpclient_head_cb)(h2o_httpclient_t *client, const char *errstr, int minor_version,
+                                                           int status, h2o_iovec_t msg, h2o_header_t *headers,
+                                                           size_t num_headers, int rlen);
+typedef h2o_httpclient_head_cb (*h2o_httpclient_connect_cb)(h2o_httpclient_t *client, const char *errstr, h2o_iovec_t *method, h2o_url_t *url,
+                                                              const h2o_header_t **headers, size_t *num_headers, h2o_iovec_t *body,
+                                                              h2o_httpclient_proceed_req_cb *proceed_req_cb, h2o_httpclient_features_t features, h2o_url_t *origin);
+typedef int (*h2o_httpclient_informational_cb)(h2o_httpclient_t *client, int minor_version, int status, h2o_iovec_t msg,
+                                                h2o_header_t *headers, size_t num_headers);
+
+typedef struct st_h2o_httpclient_connection_pool_t {
```
kazuho: Assuming that my understanding is true, I would appreciate it if you could add a doc-comment stating that the socketpool being referred to is an object that is shared among multiple threads whereas this object is thread-local.

I think having that stated would help people reading the code.

include/h2o/httpclient.h
```diff
@@ -0,0 +1,109 @@
...
+     */
+    h2o_socketpool_t *socketpool;
+
+    struct {
+        h2o_linklist_t conns;
+    } http2;
+
+} h2o_httpclient_connection_pool_t;
+
+typedef struct st_h2o_httpclient_ctx_t {
```
kazuho: I would also appreciate it if you could clarify the role of the object as well as why it is a separate structure from `h2o_httpclient_connection_pool_t`.

My understanding is that they are separate because this ctx_t can be for example path-specific (having different io-timeout for every path). Am I correct?

lib/common/http1client.c
```diff
@@ -25,53 +25,18 @@
...
-static void close_client(struct st_h2o_http1client_private_t *client)
-{
-    if (client->super.sock != NULL) {
-        if (client->super.sockpool.pool != NULL && client->_do_keepalive) {
+#include "h2o/httpclient_internal_h1.h"
+#include "h2o/token.h"
 
+static void close_client(struct st_h2o_http1client_t *client)
+{
+    if (client->sock != NULL) {
```
kazuho: Do we still need this `if` now that we construct http1client after the connection is established?

include/h2o/httpclient_internal_h1.h
```diff
@@ -0,0 +1,65 @@
...
+    } _body_decoder;
+    h2o_socket_cb reader;
+    h2o_httpclient_proceed_req_cb proceed_req;
+    char _chunk_len_str[(sizeof(H2O_UINT64_LONGEST_HEX_STR) - 1) + 2 + 1]; /* SIZE_MAX in hex + CRLF + '\0' */
+    h2o_buffer_t *_body_buf;
+    h2o_buffer_t *_body_buf_in_flight;
+    unsigned _is_chunked : 1;
+    unsigned _body_buf_is_done : 1;
+
+    h2o_mem_pool_t pool;
```
kazuho: Can't we use the pool of the request that originated the request? We have been doing that, and I do not think that we would want to start having our own pool for two reasons:
* creating a pool for every request is an additional overhead
* this would be source of bugs, because all the objects allocated from this pool and passed to output stream will become inaccessible when the http1client is being disposed

lib/common/http1client.c
i110: yes, I think it's needed because there's a possibility that `steal_socket` is called, which set client->socket NULL, before `close_client` is called.

include/h2o/httpclient.h
```diff
@@ -0,0 +1,111 @@
...
+} h2o_httpclient_properties_t;
+
+typedef void (*h2o_httpclient_proceed_req_cb)(h2o_httpclient_t *client, size_t written, int is_end_stream);
+typedef int (*h2o_httpclient_body_cb)(h2o_httpclient_t *client, const char *errstr);
+typedef h2o_httpclient_body_cb (*h2o_httpclient_head_cb)(h2o_httpclient_t *client, const char *errstr, int minor_version,
+                                                           int status, h2o_iovec_t msg, h2o_header_t *headers,
+                                                           size_t num_headers, int rlen, int header_requires_dup);
+typedef h2o_httpclient_head_cb (*h2o_httpclient_connect_cb)(h2o_httpclient_t *client, const char *errstr, h2o_iovec_t *method, h2o_url_t *url,
+                                                              const h2o_header_t **headers, size_t *num_headers, h2o_iovec_t *body,
+                                                              h2o_httpclient_proceed_req_cb *proceed_req_cb, h2o_httpclient_properties_t props, h2o_url_t *origin);
```
kazuho: I'd assume that `props` can be passed as a pointer?

i110: @kazuho fixed at https://github.com/h2o/h2o/pull/1549/commits/3f6416d39928773ff1be24ec054d94abd65982a9

kazuho:
@i110 Would you mind fixing the conflicts?

i110:
@kazuho done

lib/common/httpclient.c
```diff
@@ -0,0 +1,199 @@
...
+        assert(errstr != NULL);
+        on_connect_error(client, errstr);
+        return;
+    }
+
+    h2o_iovec_t alpn_proto;
+    if (sock->ssl == NULL || (alpn_proto = h2o_socket_ssl_get_selected_protocol(sock)).len == 0) {
+        h2o_http1client_on_connect(client, sock, origin);
+    } else {
+        if (memcmp(alpn_proto.base, "h2", alpn_proto.len) == 0) {
```
kazuho: Please use `h2o_memis`, unless we can be very sure that `alpn_proto.len` is longer than two octets. Do we have such an assumption?

lib/http2/stream.c
```diff
@@ -217,6 +217,13 @@ static int is_blocking_asset(h2o_req_t *req)
 
 static int send_headers(h2o_http2_conn_t *conn, h2o_http2_stream_t *stream)
 {
+    if (stream->req.upstream_refused) {
+        h2o_http2_encode_rst_stream_frame(&conn->_write.buf, stream->stream_id, -H2O_HTTP2_ERROR_INTERNAL);
```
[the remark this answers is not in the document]
i110: fixed at https://github.com/h2o/h2o/pull/1549/commits/ab321423443709dc3d463bb109b7c9f49a488f69

lib/common/httpclient.c
i110: fixed at https://github.com/h2o/h2o/pull/1549/commits/e3b5ab0dc8e3b3b2b0af4427742a7cd21c59e37a

i110:
@kazuho Thank you for the comments. I addressed the issues.

include/h2o/http2_common.h
```diff
@@ -0,0 +1,294 @@
...
+                                    size_t max_frame_size, const h2o_url_scheme_t *scheme, h2o_iovec_t authority,
+                                    h2o_iovec_t method, h2o_iovec_t path, h2o_headers_t *headers, uint32_t parent_stream_id);
+void h2o_hpack_flatten_response(h2o_buffer_t **buf, h2o_hpack_header_table_t *header_table, uint32_t stream_id,
+                                size_t max_frame_size, int status, h2o_headers_t headers, const h2o_iovec_t *server_name,
+                                size_t content_length);
+void h2o_hpack_flatten_request(h2o_buffer_t **buf, h2o_hpack_header_table_t *header_table, uint32_t stream_id,
+                               size_t max_frame_size, h2o_iovec_t method, h2o_url_t *url, h2o_headers_t *headers,
+                               int is_end_stream);
+void h2o_hpack_flatten_trailers(h2o_buffer_t **buf, h2o_hpack_header_table_t *header_table, uint32_t stream_id,
+                                size_t max_frame_size, h2o_header_t *headers, size_t num_headers);
```
kazuho: The "flatten" functions takes a immutable list of headers as the argument. Considering that, I think that the way of passing the headers should be by using a pointer to the list (i.e. array) and the number of items for maximum flexibility, instead of using `h2o_headers_t *` or `h2o_headers_t`.

lib/handler/proxy.c
```diff
@@ -60,38 +65,51 @@ static void on_context_init(h2o_handler_t *_self, h2o_context_t *ctx)
...
 static void on_context_dispose(h2o_handler_t *_self, h2o_context_t *ctx)
 {
     struct rp_handler_t *self = (void *)_self;
-    h2o_http1client_ctx_t *client_ctx = h2o_context_get_handler_context(ctx, &self->super);
+    struct rp_handler_context_t *handler_ctx = h2o_context_get_handler_context(ctx, &self->super);
 
-    if (client_ctx == NULL)
-        return;
+    if (handler_ctx->client_ctx != NULL) {
+        h2o_socketpool_unregister_loop(self->sockpool, ctx->loop);
```
kazuho: This might be a bug in master, but would you mind looking into if we should call `h2o_socketpool_unregister_loop` regardless of the value of `client_ctx`?

examples/libh2o/http1client.c
```diff
@@ -39,64 +39,60 @@ static h2o_iovec_t iov_filler;
...
 
-static h2o_http1client_head_cb on_connect(h2o_http1client_t *client, const char *errstr, h2o_iovec_t **reqbufs, size_t *reqbufcnt,
-                                          int *method_is_head, h2o_http1client_proceed_req_cb *proceed_req_cb,
-                                          h2o_iovec_t *cur_body, int *body_is_chunked, h2o_url_t *origin);
-static h2o_http1client_body_cb on_head(h2o_http1client_t *client, const char *errstr, int minor_version, int status,
-                                       h2o_iovec_t msg, h2o_header_t *headers, size_t num_headers, int rlen);
+static h2o_httpclient_head_cb on_connect(h2o_httpclient_t *client, const char *errstr, h2o_iovec_t *method, h2o_url_t *url,
+                                         const h2o_header_t **headers, size_t *num_headers, h2o_iovec_t *body,
+                                         h2o_httpclient_proceed_req_cb *proceed_req_cb, h2o_httpclient_properties_t props,
+                                         h2o_url_t *origin);
```
kazuho: I see the compiler complaining about type mismatches. Would you mind looking into the issue?

include/h2o.h
```diff
@@ -1968,6 +1851,7 @@ typedef struct st_h2o_proxy_config_vars_t {
     uint64_t io_timeout;
     uint64_t connect_timeout;
     uint64_t first_byte_timeout;
+    uint64_t keepalive_timeout;
```
kazuho: Is the intent here to have a timeout specific to HTTP/2? I ask this because IIUC timeout of HTTP/1 client would be handled using the timeout of HTTP/1.

If that is the case, would it make sense to rename it to h2_keepalive_timeout?

include/h2o/httpclient_internal_h2.h
```diff
@@ -0,0 +1,115 @@
...
+    h2o_url_t origin_url;
+    h2o_socket_t *sock;
+    enum enum_h2o_http2client_conn_state state;
+    khash_t(stream) * streams;
+    h2o_linklist_t link;
+    h2o_http2_settings_t peer_settings;
+    uint32_t max_open_stream_id;
+    size_t num_streams;
+    h2o_timer_t io_timeout;
+    h2o_timer_t keepalive_timeout;
```
kazuho: Can't we use one timer object instead of having two?

My understanding is that their use is exclusive; i.e. keepalive_timeout is unset when io_timeout is used, io_timeout is unset when keepalive_timeout is used.

include/h2o/http2_common.h
```diff
@@ -0,0 +1,294 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+#ifndef h2o__foobar_h
+#define h2o__foobar_h
```
kazuho: Please change this to an appropriate name.

include/h2o/httpclient_internal_h1.h
```diff
@@ -0,0 +1,63 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+#ifndef h2o__http1client_h
+#define h2o__http1client_h
```
kazuho: Maybe add `_internal`?

include/h2o.h
i110: No, this is also used for HTTP/1 at https://github.com/h2o/h2o/blob/1d602e58f2aaf864cd3900a9ccf634910856f522/lib/handler/configurator/proxy.c#L371

kazuho: Ah! Thank you for the clarification.

lib/handler/proxy.c
kazuho: fixed in #1851.

kazuho:
Congratulations! The branch has been merged.

While the multi-protocol support looks fine, my understanding is that the HTTP/2 client implementation is still WIP. Let's track and address the issues using #1852.

PR #1807, opened 2018-07-11
Invalid status json with multiple status endpoints

cho45:
Following config produce invalid duplicated json:

```
hosts:
  default:
    paths:
      /server-status:
        status: ON
      /server-status2:
        status: ON
```

### DIFF
--- a/lib/handler/status.c
+++ b/lib/handler/status.c
@@ -265,9 +265,10 @@ void h2o_status_register(h2o_pathconf_t *conf)
     self->super.on_context_init = on_context_init;
     self->super.on_context_dispose = on_context_dispose;
     self->super.on_req = on_req;
-    if (!handler_registered++) {
+    if (!handler_registered) {
+        handler_registered++;
         h2o_config_register_status_handler(conf->global, requests_status_handler);
         h2o_config_register_status_handler(conf->global, events_status_handler);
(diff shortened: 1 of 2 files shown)


lib/handler/status.c
```diff
@@ -260,11 +260,15 @@ static void on_context_dispose(h2o_handler_t *_self, h2o_context_t *ctx)
...
 {
+    static int handler_registered = 0;
     struct st_h2o_root_status_handler_t *self = (void *)h2o_create_handler(conf, sizeof(*self));
     self->super.on_context_init = on_context_init;
     self->super.on_context_dispose = on_context_dispose;
     self->super.on_req = on_req;
-    h2o_config_register_status_handler(conf->global, requests_status_handler);
-    h2o_config_register_status_handler(conf->global, events_status_handler);
-    h2o_config_register_status_handler(conf->global, durations_status_handler);
+    if (!handler_registered) {
```
deweerdt: Do you think it's worth erroring out? It's likely that this is a configuration error that the user might want to know about?

cho45: I think this case shoud not raise an error.
For my case, I want to serve same status for an internal domain and an external domain with access control.

deweerdt: Sorry, i meant the `else` case for this: otherwise we're simply silently ignoring invalid configuration?

PR #1834, opened 2018-08-17
[doc] fix incorrecet example in doc

kazan417:
correcting code for restricting access to 192.168. Old code did opposite.
If i correctly understand ruby code It looks like code in example did opposite than explanation for this code:
By using the 399 status code, it is possible to implement access control using mruby. The example below restricts access to requests from 192.168. private address.

### DIFF
--- a/doc/configure/mruby.html
+++ b/doc/configure/mruby.html
@@ -109,9 +109,9 @@ The example below restricts access to requests from <code>192.168.</code> privat
     mruby.handler: |
       lambda do |env|
         if /\A192\.168\./.match(env[&quot;REMOTE_ADDR&quot;])
-          return [399, {}, []]
+          return [403, {&#39;content-type&#39; =&gt; &#39;text/plain&#39;}, [&quot;access forbidden\n&quot;]]          
         end
-        [403, {&#39;content-type&#39; =&gt; &#39;text/plain&#39;}, [&quot;access forbidden\n&quot;]]
+        return [399, {}, []]
       end
 </code></pre>
 </div>


kazan417:
in some examples we have just [399, headers, []] without "return" word before it. Does ruby handle it?

proyb6:
From what I understand your PR would block access to 192.168.x.x in contrast to the original documentation which whitelisting/restricting access to 192.168.x.x

For me, ```return [403, {'content-type' => 'text/plain'}, ["access forbidden\n"]]``` is correct and less confuse.

MRuby Lambdas vs Proc can be rather confuse for some and me who aren't good in (M)Ruby in the beginning.

In ```Proc.new``` aka ```proc```, you cannot declare ```return``` keyword or it will throw an error that appear in your shell/console, the proc will have it own return that return result.

proyb6:
From what I understand your PR would block access to 192.168.x.x in contrast to the original documentation which whitelisting/restricting access to 192.168.x.x

MRuby Lambdas vs Proc can be rather confuse for some and me who aren't good in (M)Ruby in the beginning.

In ```Proc.new``` aka ```proc```, you cannot declare ```return``` keyword or it will throw an error that appear in your shell/console, the proc will have it own return that return result.

proyb6:
From what I understand your PR would block access to 192.168.x.x in contrast to the original documentation which whitelisting/restricting access to 192.168.x.x

MRuby Lambdas vs Proc can be rather confuse for some and me who aren't good in (M)Ruby in the beginning.

In ```Proc.new``` aka ```proc```, you cannot declare ```return``` keyword or it will throw an error that appear in your shell/console, the proc will have it own return that return result.

In Lambdas, you can declare ```return``` which is useful if you need to compare incoming values/data against your value and more easier to reason about.

i110:
I think the current document is correct, which allows only `192.168.*` IPs to access the resources. (Though I think it's nice if we can rewrite that sentence using more easy expression)

> in some examples we have just [399, headers, []] without "return" word before it. Does ruby handle it?

yes, without explicit return (m)ruby uses the return value of last evaluated expression as a return value of lambda / method.

BTW if you want to refine the docs, please don't modify `doc/*.html` directly. see https://github.com/h2o/h2o/wiki/Editing-and-compiling-Documentation

kazan417:
Now i understand what this words means, but "google translate"  can not.
 maybe is better to say:
The example bellow restricts access for requests by allowing access for addresses that begin with 192.168. only. 
Is is little bit better in my oppinion.

kazan417:
Move to other pull request that fixes explanation without touching example code 
https://github.com/h2o/h2o/pull/1836

kazan417:
#1836

PR #1890, opened 2018-10-30
Connection culling

deweerdt:
This PR implements a mechanism to cleanup idle connections when we reach `max-connections` (as opposed to delaying `accept(2)`).

### DIFF
--- a/lib/core/context.c
+++ b/lib/core/context.c
@@ -93,8 +93,8 @@ void h2o_context_init(h2o_context_t *ctx, h2o_loop_t *loop, h2o_globalconf_t *co
     h2o_multithread_register_receiver(ctx->queue, &ctx->receivers.hostinfo_getaddr, h2o_hostinfo_getaddr_receiver);
     ctx->filecache = h2o_filecache_create(config->filecache.capacity);
 
-    h2o_linklist_init_anchor(&ctx->http1._conns);
-    h2o_linklist_init_anchor(&ctx->http2._conns);
+    h2o_linklist_init_anchor(&ctx->_active_conns);
+    h2o_linklist_init_anchor(&ctx->_inactive_conns);
     ctx->proxy.client_ctx.loop = loop;
     ctx->proxy.client_ctx.io_timeout = ctx->globalconf->proxy.io_timeout;
     ctx->proxy.client_ctx.connect_timeout = ctx->globalconf->proxy.connect_timeout;
@@ -186,4 +186,24 @@ void h2o_context_update_timestamp_string_cache(h2o_context_t *ctx)
     gmtime_r(&ctx->_timestamp_cache.tv_at.tv_sec, &gmt);
     h2o_time2str_rfc1123(ctx->_timestamp_cache.value->rfc1123, &gmt);
     h2o_time2str_log(ctx->_timestamp_cache.value->log, ctx->_timestamp_cache.tv_at.tv_sec);
+}
+
+int h2o_context_close_idle_connections(h2o_context_t *ctx, int max_connections_to_close)
+{
+    int closed = 0;
+    h2o_linklist_t *node, *nprev;
+    for (node = ctx->_inactive_conns.prev; node != &ctx->_inactive_conns; node = nprev) {
+        struct timeval now;
+        h2o_conn_t *conn = H2O_STRUCT_FROM_MEMBER(h2o_conn_t, _conns, node);
+        nprev = node->prev;
+
+        now = h2o_gettimeofday(ctx->loop);
+        if (conn->connected_at.tv_sec == now.tv_sec && conn->connected_at.tv_usec == now.tv_usec)
+            continue;
+        if (conn->callbacks->close_idle_connection(conn))
+            closed++;
+        if (closed == max_connections_to_close)
+            break;
+    }
+    return closed == max_connections_to_close;
 }
--- a/lib/http1.c
+++ b/lib/http1.c
@@ -54,7 +54,6 @@ struct st_h2o_http1_conn_t {
     h2o_conn_t super;
     h2o_socket_t *sock;
     /* internal structure */
-    h2o_linklist_t _conns;
     h2o_timer_t _timeout_entry;
     uint64_t _req_index;
     size_t _prevreqlen;
@@ -131,7 +130,7 @@ static void close_connection(struct st_h2o_http1_conn_t *conn, int close_socket)
     h2o_dispose_request(&conn->req);
     if (conn->sock != NULL && close_socket)
         h2o_socket_close(conn->sock);
-    h2o_linklist_unlink(&conn->_conns);
+    h2o_linklist_unlink(&conn->super._conns);
     free(conn);
 }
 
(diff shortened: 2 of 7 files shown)


lib/http1.c
```diff
@@ -994,15 +1008,27 @@ static h2o_iovec_t log_request_index(h2o_req_t *req)
     return h2o_iovec_init(s, len);
 }
 
+static int conn_is_h1(h2o_conn_t *conn)
+{
+    return conn->callbacks->get_sockname == get_sockname;
```
kazuho: How about promoting `static const h2o_conn_callbacks_t callbacks` defined in `h2o_http1_accept` to a file-level static variable and comparing its address with `conn->callbacks`?

Comparing the address of the callback table is more natural than comparing a particular value of a callback table.

lib/http1.c
```diff
@@ -529,8 +544,7 @@ static void reqread_on_timeout(h2o_timer_t *entry)
         conn->req.res.reason = "Request Timeout";
     }
 
-    conn->req.http1_is_persistent = 0;
-    close_connection(conn, 1);
+    close_idle_connection(&conn->super);
```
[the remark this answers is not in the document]
deweerdt: That's a good point. I do assume that `conn->sock->input->size == 0` because `reqread_on_timeout` fired, which means there was nothing to read and we ended up timing out. I'll add the assert.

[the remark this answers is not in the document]
kazuho: Don't we see a timeout with a non-empty `conn->sock->input` if the client has sent a partial HTTP2 frame?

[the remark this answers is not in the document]
deweerdt: > Don't we see a timeout with a non-empty conn->sock->input if the client has sent a partial HTTP2 frame?

Would that happen in http1? In a connection upgrade?

deweerdt:
> Would it be preferable to start closing idle connections before we reach max_connections?

Yes, you're correct, it would. I was trying to avoid a soft limit, but i realize now that there's no escaping.

> Would it make sense to guarantee that idle connections would be kept open for a small amount of seconds? 

Yes, it would. Should we hardcode the value to, say, 30s, or would you see a use case to make this configurable?

It sounds like we would end up with one or two additional parameters:
- `soft-connection-limit: <number>`: a number lower than `max-connections` that would have H2O start closing idle connections?
- `soft-connection-limit.min-age: <duration>`: the minimum amount of time that we should wait before considering a connection idle.

deweerdt:
@i110 I believe that edf760b addresses the comments you made so far. I will address the suggestions that @kazuho had.

src/main.c
```diff
@@ -1565,13 +1565,16 @@ static void on_accept(h2o_socket_t *listener, const char *err)
...
-        if (num_connections(0) >= conf.max_connections) {
+        int excess_connections = num_connections(0) - conf.max_connections;
+        if (excess_connections >= 0) {
             /* The accepting socket is disactivated before entering the next in `run_loop`.
              * Note: it is possible that the server would accept at most `max_connections + num_threads` connections, since the
              * server does not check if the number of connections has exceeded _after_ epoll notifies of a new connection _but_
              * _before_ calling `accept`.  In other words t/40max-connections.t may fail.
              */
-            break;
+            if (!h2o_context_close_idle_connections(ctx->accept_ctx.ctx, excess_connections)) {
```
i110: I still see two issues in this line:

1. we should pass `excess_connections + 1` to ensure that there will be at least one slot for the incoming connection
2. for now `h2o_context_close_idle_connections` returns the number of closed connections, so we should compare that with `excess_connections + 1`

Thus I think the following is the correct one:
```c
if (h2o_context_close_idle_connections(ctx->accept_ctx.ctx, excess_connections + 1) < excess_connections + 1) {
```

i110: And also I realized that it's dangerous to pass `0` to `h2o_context_close_idle_connections` mistakenly like this case, because it will try to close all idle connections if we do so. How about adding `assert(max_connections_to_close > 0)` in the beginning of that function?

deweerdt:
@i110 @kazuho I believe that 5f46d7e addresses all your comments so far. In introduces `soft-connection-limit`, which set limits at which H2O will try to close connections older than `soft-connection-limit-min-age`seconds

src/main.c
```diff
@@ -1565,14 +1579,21 @@ static void on_accept(h2o_socket_t *listener, const char *err)
 
     do {
         h2o_socket_t *sock;
-        if (num_connections(0) >= conf.max_connections) {
+        if (conf.soft_connection_limit > 0) {
+            int excess_connections = num_connections(0) - conf.soft_connection_limit;
```
kazuho: Should we better divide the number with `conn.num_threads`? Otherwise I think we'd be retiring connections unfairly aggressively on the thread that hits this condition first.

src/main.c
```diff
@@ -1565,14 +1579,21 @@ static void on_accept(h2o_socket_t *listener, const char *err)
 
     do {
         h2o_socket_t *sock;
-        if (num_connections(0) >= conf.max_connections) {
+        if (conf.soft_connection_limit > 0) {
```
kazuho: I think we might want to set `conf.soft_connection_limit` to `MAX_INT` to avoid this `if`.

lib/core/context.c
```diff
@@ -187,3 +187,25 @@ void h2o_context_update_timestamp_string_cache(h2o_context_t *ctx)
...
     h2o_time2str_log(ctx->_timestamp_cache.value->log, ctx->_timestamp_cache.tv_at.tv_sec);
 }
+
+int h2o_context_close_idle_connections(h2o_context_t *ctx, int max_connections_to_close, int min_age)
+{
+    int closed = 0;
+    h2o_linklist_t *node, *nprev;
+    if (max_connections_to_close <= 0)
+        return 0;
+    for (node = ctx->_inactive_conns.prev; node != &ctx->_inactive_conns; node = nprev) {
```
kazuho: Shouldn't we iterate the list in the normal order (i.e. by using `next`)?

I assume that the elements are ordered from the oldest to the newest, because we insert connections like by doing `h2o_linklist_insert(&conn->super.ctx->_inactive_conns, &conn->super._conns)`, which inserts the element at the end of the list. Closing the ones that have spent longer time in idle state makes sense to me.

lib/core/context.c
```diff
@@ -187,3 +187,25 @@ void h2o_context_update_timestamp_string_cache(h2o_context_t *ctx)
...
+        struct timeval now;
+        h2o_conn_t *conn = H2O_STRUCT_FROM_MEMBER(h2o_conn_t, _conns, node);
+        nprev = node->prev;
+
+        now = h2o_gettimeofday(ctx->loop);
+        if (now.tv_sec - conn->connected_at.tv_sec < min_age)
+            continue;
```
kazuho: We can do a break here, assuming that we make the change above.

That is because a connection that has lasted shorter than min_age is guaranteed to have spent less than that amount of time in the idle state.

lib/http1.c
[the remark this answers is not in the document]
kazuho: My bad, please ignore my comment here.

Issue #1929, opened 2018-12-14
Shall websocket client be supported for libh2o, just like server side websocket?


chenbd:
I'll issue a pull request for this If this sounds good.

meritozh:
Great! I'd like to see. By the way, if you have any time, would you do me a favor? #1926 I'm reading h2o source code and want do some contributions in the future.

PR #1931, opened 2018-12-15
add websocket client for libh2o

chenbd:
According to issue #1929


shazni:
This looks like committed but I can't seem to find the websocketclient.c

If it by chance not exists, it would be great to add a websocketclient example to libh2o with SSL support.

PR #1966, opened 2019-02-09
Update mruby to latest (>2.0.0) version

chidea:
Some of mruby gems started to loose its support for lower version of mruby at its latest versions.
So I decided to upgrade mruby in h2o to the latest version.

While it was easier than I thought to do this, there were several decisions could be made.
1. `lib/handler/mruby/middleware.c` has `create_subreq` function and it was utilizing khash table library to iterate through table items. With latest mruby, there are new MRB_API function called `mrb_hash_foreach` which does the same. It could be used to reduce khash library usage on it and reduce this somewhat long function. However, a downside here is that the foreach function is introduced after mruby 2.0.0 stable version. I've included the latest version and edited some lines in `middleware.c` in this PR, with a struct `on_mrb_hash_foreach_arg` to pass arguments between them.

2. `deps/mruby-require` had to be updated to the latest `iij/mruby-require`. As mentioned before, this library drops support on old versions of mruby.

3. `deps/mruby-pack` had to be replaced. This is now included in mruby by default since 2.0.0 stable.

4. I tested `zzak/mruby-uri` to use URI www/html encoding/decoding in form body. I don't understand why this kind of library isn't included before, but for a webserver framework, this seemed pretty decent to be. After some tests, this library was proved to be impossible to be placed in `deps`. Some ways around were either adding a line `conf.gem :github => 'zzak/mruby-uri', :branch => 'master'` into `misc/mruby_config.rb` or just git cloning into `deps/mruby/mrbgems`. I guess this kind of function had to be included in original h2o library and some embedded code in mruby had to provide URI module to call it instead, but h2o actually provides only one-sided functions like `htmlescape`.

It's okay if you deny this pull request until the stable mruby gets out since I understand mruby things aren't on the first priority on your list and it may fail some tests. I'm just leaving it to relief your work when mruby has newer version.


chidea:
5. `.gitignore` has `mruby/` which was also making git to ignore any changes in `deps/mruby`. Had to change it to `./mruby/`

i110:
I filed https://github.com/h2o/h2o/pull/1970 to replace this, because this needs some fixes in `middleware.c` and also we have [a policy](https://github.com/h2o/h2o/wiki/Importing-third-party-code) to include third party libraries into h2o code base. But anyway this was a great help, so thank you so much!

PR #1979, opened 2019-02-21
[http2 client] Make sure the connection is registered with the stream before calling on_connect

deweerdt:
This is helpful in case the on_connect callback is interested in
examining the connection in question.


i110:
@deweerdt thank you for the PR. I think the change itself is okay, but just out of curiosity, what kind of examining do you want to do? I'm asking because there is `h2o_httpclient_properties_t` that can be used to pass information from httpclient to the caller side.

PR #2050, opened 2019-05-11
http2 upgrade of POST request

i110:
[RFC7540 3.2](https://tools.ietf.org/html/rfc7540#section-3.2) allows us to do http2 upgrade even if the request has the body. 

Before this PR,
1. a request with body payload is never upgraded to http2
2.  `OPTIONS` upgrade request is processed too (i.e. the client gets 2 responses)

This PR make h2o able to upgrade such a request to http2 too, and stop processing `OPTIONS` upgrade request.

### DIFF
--- a/lib/http1.c
+++ b/lib/http1.c
@@ -148,13 +148,18 @@ static void set_timeout(struct st_h2o_http1_conn_t *conn, uint64_t timeout, h2o_
         h2o_timer_link(conn->super.ctx->loop, timeout, &conn->_timeout_entry);
 }
 
-static void process_request(struct st_h2o_http1_conn_t *conn)
+static int should_upgrade_to_http2(struct st_h2o_http1_conn_t *conn)
 {
-    if (conn->sock->ssl == NULL && conn->req.upgrade.base != NULL && conn->super.ctx->globalconf->http1.upgrade_to_http2 &&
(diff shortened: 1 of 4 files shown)


lib/http1.c
```diff
@@ -540,7 +545,7 @@ static void handle_incoming_request(struct st_h2o_http1_conn_t *conn)
             if (create_entity_reader(conn, headers + entity_body_header_index) != 0) {
                 return;
             }
-            conn->req.write_req.cb = h2o_write_req_first;
+            conn->req.write_req.cb = should_upgrade_to_http2(conn) ? write_req_non_streaming : h2o_write_req_first;
```
i110: I think this is controversial. This current implementation prefers upgrade to body streaming, but it might be better to rather select body streaming on h1 regarding the performance. What do you think @kazuho ?

i110:
> OPTIONS upgrade request is processed too (i.e. the client gets 2 responses)

This was my misunderstanding. h2o should send the response for OPTIONS request too (and it's via h2), so the master's behavior is not wrong.

Actually almost all of my motivation of this PR was mainly fixing above (for testing), because http2 upgrade is not being frequently used nowadays thanks to TLS, especially with body payload. So I close this PR and will open again when we need this in the future.

PR #2171, opened 2019-11-16
Quic ptls leak

deweerdt:
the picotls members of the `quicly_conn_t` struct wouldn't be free'd when the connection would be disposed of, resulting in a relatively large per connection leak.


deweerdt:
I just realized i should send post this to the quicly repo instead.

deweerdt:
https://github.com/h2o/quicly/pull/223

PR #2217, opened 2019-12-23
[http2] Add initial absolute priority header support

hfujita:
This PR adds an initial absolute priority header support based on [draft-kazuho-httpbis-priority-04](https://tools.ietf.org/html/draft-kazuho-httpbis-priority-04).

This PR is an initial prototype and has several assumptions:

- Back-end (e.g. file handler, reverse proxy origin) adds the priority header.
- Request priorities from a client are modified based on back-end priority headers.
- As for client-generated priorities, only assumes Chromium-style dependency tree (linear, weight-sorted list)
- Only `urgency` parameter is interpreted. `progressive` parameter is ignored at the moment.
- Only priority promotion is supported (decreasing priority is not supported)


hfujita:
Fixed code formatting using `clang-format`.

lib/http2/stream.c
```diff
@@ -271,12 +272,40 @@ static int send_headers(h2o_http2_conn_t *conn, h2o_http2_stream_t *stream)
         if (is_blocking_asset(&stream->req))
             h2o_http2_scheduler_rebind(&stream->_scheduler, &conn->scheduler, 257, 0);
     } else {
-        /* raise the priority of asset files that block rendering to highest if the user-agent is _not_ using dependency-based
-         * prioritization (e.g. that of Firefox)
-         */
-        if (conn->num_streams.priority.open == 0 && stream->req.hostconf->http2.reprioritize_blocking_assets &&
-            h2o_http2_scheduler_get_parent(&stream->_scheduler) == &conn->scheduler && is_blocking_asset(&stream->req))
+        /* Handle absolute priority header */
+        ssize_t absprio_cursor = h2o_find_header_by_str(&stream->req.res.headers, H2O_STRLIT("priority"), -1);
```
deweerdt: it might be worth adding `priority` to the tokens: https://github.com/h2o/h2o/blob/master/misc/tokens.pl

deweerdt:
I've noticed CI fails in `t/50origin-frame.t`, this is because there are two SETTINGS frames sent now. I think this isn't necessary, since a single SETTINGS frame can contain two settings, the following should work:
```
 static const uint8_t SERVER_PREFACE_BIN[] = {
     /* settings frame */
-    LIT_FRAME_HEADER(6, H2O_HTTP2_FRAME_TYPE_SETTINGS, 0, 0), LIT16(H2O_HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS), LIT32(100),
-    LIT_FRAME_HEADER(6, H2O_HTTP2_FRAME_TYPE_SETTINGS, 0, 0), LIT16(H2O_HTTP2_SETTINGS_DEPRECATE_HTTP2_PRIORITIES), LIT32(1),
+    LIT_FRAME_HEADER(12, H2O_HTTP2_FRAME_TYPE_SETTINGS, 0, 0), LIT16(H2O_HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS), LIT32(100),
+                                                               LIT16(H2O_HTTP2_SETTINGS_DEPRECATE_HTTP2_PRIORITIES), LIT32(1),
     /* window_update frame */
```

hfujita:
@deweerdt Thanks for the feedback! I'll address the things you pointed out and get back to you shortly.

As for the design, I'm more than happy for suggestions that makes this more robust against various clients. I have some questions regarding your idea though:
> At the creation of the connection, H2O reservers placeholders in the scheduler, one for each urgency, with the appropriate weight.

I'm not sure if I understand this correctly. How are these placeholders placed within the scheduler dependency tree? Are they in parent-child relationships? (e.g. `placeholder-for-u0` is a parent of `placeholder-for-u1` in the initial state?)

include/h2o/absprio.h
```diff
@@ -0,0 +1,46 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+#ifndef h2o__absprio_h
+#define h2o__absprio_h
+
```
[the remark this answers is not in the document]
hfujita: Hmm. I'm not sure if this is the case. I would think this is still an internal API that will be commonly used by both http2 and http3. I think `http2_scheduler.h` is close to this situation, which doesn't have `extern "C"` either. https://github.com/h2o/h2o/blob/master/include/h2o/http2_scheduler.h

deweerdt:
> I'm not sure if I understand this correctly. How are these placeholders placed within the scheduler dependency tree? Are they in parent-child relationships? (e.g. placeholder-for-u0 is a parent of placeholder-for-u1 in the initial state?)

They'd be siblings. Today, we have a root, under which all streams are prioritized by the browser. How about a root under which we have siblings with different weights, and the `H2O_ABSPRIO_URGENCY_DEFAULT` sibling gets all streams prioritized by the browser (ie, it's stream 0 wrt to http/2 priorities), and when the server side prioritizes, that moves the stream to one of the ABSPRIO siblings.

include/h2o/absprio.h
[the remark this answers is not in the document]
deweerdt: ah, fair enough

deweerdt:
Here's a drawing of today's implementation vs the proposal i made: image

lib/common/absprio.c
```diff
@@ -0,0 +1,120 @@
...
+     */
+    enum en_absprio_parser_state_t state = EXPECT_KEY;
+    size_t c = 0;
+    const char *s = value->base;
+    char key;
+
+    while (c < value->len) {
+        switch (state) {
+        case EXPECT_KEY:
+            if (!isalpha(s[c]) || !islower(s[c])) {
```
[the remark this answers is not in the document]
kazuho: Yeah. For the purpose of this PR, I think that an alternative approach would be to use the existing `h2o_next_token` function for parsing the header field. Examples can be found in t/00unit/lib/common/string.c.

Eventually we need to have a generic parser for Structured Headers that would look like `h2o_absprio_parse_priority`, the drafts (both Structured Headers and Priority) are not finalized yet, and we can delay the hard work until later time.

lib/common/absprio.c
```diff
@@ -0,0 +1,120 @@
...
+                    break;
+                }
+                /* Unknown key */
+            }
+            /* Unexpected character -- skip until ',' or end of string */
+            state = SKIP_UNTIL_SPACE;
+            break;
+
+        case EXPECT_DIGIT:
+            if (isdigit(s[c]))
```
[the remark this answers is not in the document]
kazuho: Yeah. For the purpose of this PR, I think that an alternative approach would be to use the existing `h2o_next_token` function for parsing the header field. Examples can be found in t/00unit/lib/common/string.c.

Eventually we need to have a generic parser for Structured Headers that would look like `h2o_absprio_parse_priority`, the drafts (both Structured Headers and Priority) are not finalized yet, and we can delay the hard work until later time.

kazuho:
@deweerdt Would you mind clarifying how that would work with Chrome?

Consider the case where the server sends `Priority: u=2`. Then, h2o has to serve that response before other images, but after CSS, JS, HTML. I think the only way to do that with Chrome is to alter the "priority chain" that Chrome has built.

lib/http2/connection.c
```diff
@@ -285,6 +287,11 @@ static void preserve_stream_scheduler(h2o_http2_conn_t *conn, h2o_http2_stream_t
     h2o_http2_scheduler_deactivate(&(*dst)->_scheduler);
 }
 
+void h2o_http2_conn_preserve_stream_scheduler(h2o_http2_conn_t *conn, h2o_http2_stream_t *src)
+{
+    preserve_stream_scheduler(conn, src);
+}
```
kazuho: I think we'd prefer promoting `preserve_stream_scheduler` to an external function by adding `h2o_http2_conn_` as the prefix.

I understand the intention to minimize the changes, but that comes with the complexity of adding a wrapper, which increases the maintenance cost in the long run.

lib/http2/scheduler.c
```diff
@@ -412,3 +412,21 @@ int h2o_http2_scheduler_is_active(h2o_http2_scheduler_node_t *root)
 {
     return root->_queue != NULL && !queue_is_empty(root->_queue);
 }
+
+h2o_http2_scheduler_node_t *h2o_http2_scheduler_find_new_parent(h2o_http2_scheduler_node_t *root, uint16_t new_weight)
```
kazuho: How about renaming the function to `h2o_http2_scheduler_find_parent_by_weight` or something?

As correctly pointed out in the comment, this logic is specific to how weight is used by a particular client implementation, and I think it'd be better if the name of the function implies that.

hfujita:
@deweerdt 
>They'd be siblings.

Hmm, that's actually different from my understanding on the current proposal. IIUC, requests with lower urgency (lower `u=` value) are not getting served until requests with higher urgency are completed. In some sense this is quite similar to dependency tree where lower priority requests are descendants of higher priority requests. Making them siblings would not allow this behavior. @kazuho can you jump in and clarify?

kazuho:
@hfujita I agree with your observation, as stated in https://github.com/h2o/h2o/pull/2217#issuecomment-570747163.

lib/http2/scheduler.c
hfujita: Good point. Actually I have been struggling to come up with a better name, as this logic itself is extremely client (tree shape) dependent, but at the same time I did not want to make the name too lengthy. Let's go with `h2o_http2_scheduler_find_parent_by_weight` for now.

lib/common/absprio.c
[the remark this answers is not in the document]
hfujita: Done rewriting using `h2o_next_token`.

lib/common/absprio.c
```diff
@@ -0,0 +1,120 @@
...
+                *urgency = s[c] - '0'; /* TODO: this doesn't reject u=8,9 */
```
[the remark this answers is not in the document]
hfujita: Done.

lib/http2/scheduler.c
hfujita: Done.

lib/http2/stream.c
hfujita: Done.

lib/common/absprio.c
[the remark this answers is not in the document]
hfujita: Done.

hfujita:
@deweerdt @kazuho Updated the PR with several changes. Should be ready for another round of review.

As a result of online/offline discussions, my goal for this PR is to complete the priority header support which assumes Chromium-style dependency tree, with a verification that the tree is actually in such a shape.

FYI, here's the list of totally new commits in this push (newer in the top). I intentionally kept them separate to make the second-round review easier. Patches with `*` will be squashed with appropriate preceding patches after getting reviewed. 
```
* 5be61d6e Expand priority header test cases
* 64ee143b [http2] Support stream demotion with priority headers
* c89b47e7 [http2] Add tree-shape check on priority header handling
bcc70112 [http2] Add test cases for Chromium-shape detection
68b2293e [http2] Add Chromium-type dependency tree detection
87363ae3 [http2] Add a flag for stale scheduler entry
6f9e3b1d Add "priority" to the token table
```

include/h2o/absprio.h
```diff
@@ -0,0 +1,46 @@
...
+    H2O_ABSPRIO_URGENCY_SUPPLEMENTARY_2 = 4,
+    H2O_ABSPRIO_URGENCY_SUPPLEMENTARY_3 = 5,
+    H2O_ABSPRIO_URGENCY_SUPPLEMENTARY_4 = 6,
+    H2O_ABSPRIO_URGENCY_BACKGROUND = 7,
+} h2o_absprio_urgency_type_t;
+
+void h2o_absprio_parse_priority(const h2o_iovec_t *header_value, uint8_t *urgency, int *incremental);
+
+/* Convert urgency value (in absolute priority header) to HTTP2 weight */
+static inline uint16_t h2o_absprio_urgency_to_weight(uint8_t urgency)
```
kazuho: Should we better call this function "urgency_to_chrome_weight" or something, as it is specific to Chrome?

lib/http2/frame.c
```diff
@@ -53,6 +53,7 @@ int h2o_http2_update_peer_settings(h2o_http2_settings_t *settings, const uint8_t
             SET(MAX_CONCURRENT_STREAMS, max_concurrent_streams, 0, UINT32_MAX, 0);
             SET(INITIAL_WINDOW_SIZE, initial_window_size, 0, 0x7fffffff, H2O_HTTP2_ERROR_FLOW_CONTROL);
             SET(MAX_FRAME_SIZE, max_frame_size, 16384, 16777215, H2O_HTTP2_ERROR_PROTOCOL);
+            SET(DEPRECATE_HTTP2_PRIORITIES, deprecate_http2_priorities, 0, 1, H2O_HTTP2_ERROR_PROTOCOL);
```
kazuho: I am not sure if we need to send this settings parameter now; aren't we only recognizing the signal from the origin (sent as a header field)? I think we should not be sending this settings parameter to the client until we support accepting the header field (or whatever the priority draft would define) from the client.

lib/http2/scheduler.c
```diff
@@ -253,6 +254,7 @@ void h2o_http2_scheduler_relocate(h2o_http2_scheduler_openref_t *dst, h2o_http2_
     dst->_self_is_active = src->_self_is_active;
     dst->_queue_node._link = (h2o_linklist_t){NULL};
     dst->_queue_node._deficit = src->_queue_node._deficit;
+    dst->_is_stale = 1;
```
kazuho: I feel a bit uneasy here, because the name of the function (i.e. "relocate") or the signature does not imply that this function is making the reference "stale." We might see confusion in the future, if we modify a flag named as "stale" in a function that "relocates" an object.

That said, I understand that we might remove the bit as we change the way we detect if the client is Chromium. Or if that is not the case, I think we might be fine with having a "stale" flag as a public field in the object, letting the http2 stack directly modify the flag.

include/h2o/http2_scheduler.h
```diff
@@ -122,6 +123,28 @@ int h2o_http2_scheduler_run(h2o_http2_scheduler_node_t *root, h2o_http2_schedule
...
 int h2o_http2_scheduler_is_active(h2o_http2_scheduler_node_t *root);
 
+/**
+ * returns if the scheduler tree is linear (no node has multiple children) and
+ * the weight is monotonically decreasing (parent's weight is greater than or equal to
+ * child's)
+ *
+ * Note: Chromium's request satisfies this property
+ */
+int h2o_http2_scheduler_is_linear_monotonic(h2o_http2_scheduler_node_t *root);
```
kazuho: If we are to retain this function, I think it might better if the name included "chrome" (or "chromium").

The rationale is that what we are doing here is checking if the priority tree being build looks like it is Chrome, and the list being linear and monotonic is merely a necessary condition. We might want to also check that the weight values are the ones that would be sent by Chrome.

lib/common/absprio.c
```diff
@@ -0,0 +1,66 @@
...
+    *incremental = token[3] - '0';
+}
+
+void h2o_absprio_parse_priority(const h2o_iovec_t *value, uint8_t *urgency, int *incremental)
+{
+    h2o_iovec_t iter = *value;
+    const char *token;
+    size_t token_len;
+
+    while ((token = h2o_next_token(&iter, ',', &token_len, NULL)) != NULL) {
```
kazuho: It is possible to obtain an `h2o_iovec_t` pointing to the value, if you pass an `h2o_iovec_t *` as the fourth argument.

```c
h2o_iovec_t value;
while ((token = h2o_next_token(&iter, ',', &token_len, &value)) != NULL) {
    if (token_len == 1 && token[0] == 'u') {
        if (value.base != NULL) {
            // value should contain '0' .. '7'
        } else {
            // no corresponding value, therefore it's an error
        }
    } else if (token_len == 1 && token[0] == 'i') {
        if (value.base != NULL) {
            // value should contain '?0' or '?1'
        } else {
            // value omitted, meaning that i is true
        }
    }
}
```

lib/http2/frame.c
hfujita: Okay, it makes sense. I'll drop this patch for now.

include/h2o/http2_scheduler.h
hfujita: My intention was that "monotonic" part would cover the property on weight (weight values would decrease monotonically as you walk through the nodes from the root), thought the word "weight" was omitted from the function name due to the length.

I have been reluctant to include "chrome" or "chromium" in symbol names because I thought it was too specific, but now I rethink that maybe it might be better to simply say "chromium" rather than pretend to be generic (like "linear" or "monotonic"), as it might be less confusing.

lib/http2/scheduler.c
hfujita: Hmm, okay then how about renaming it to `is_relocated`?

> That said, I understand that we might remove the bit as we change the way we detect if the client is Chromium.

I don't think that's not the case, unfortunately. Regardless of the way we implement Chromium detection, I believe this flag is necessary anyway to implement `find_parent_by_weight` function correctly.

lib/common/absprio.c
hfujita: 👍 Was not aware of this feature. Let me incorporate your suggestion.

Also, I was totally missing out that sh-boolean value can be omitted if it's true. I'll include this in the test case as well.

include/h2o/absprio.h
hfujita: 👍 Good catch.

hfujita: Done

hfujita:
@kazuho Addressed your comments, plus implemented Chromium detection at frame reception. (Removed `is_linear_monotonic` function as a result.)

Now seeing a weird CI failure (`40server-push.t`) only in openssl-1.1.0 build. Looking into it.

hfujita:
@kazuho @deweerdt told me that the failed test (`40server-push.t`) was known to occasionally fail. I think the PR is ready for another round of review, and that mostly ready-to-go unless you find another issue. Once approved, I would like to squash some of the patches, so please give me an opportunity to do so before pushing the merge button.

include/h2o/http2_scheduler.h
hfujita: This function went away in the latest revision.

lib/common/absprio.c
kazuho: FWIW, I do not think you missed anything. Omission of "=?1" is a recent change to Structured Headers :-)

include/h2o/http2_internal.h
```diff
@@ -164,6 +164,12 @@ struct st_h2o_http2_conn_t {
     /* internal */
     h2o_http2_scheduler_node_t scheduler;
     h2o_http2_conn_state_t state;
+    int chromium_dependency_tree; /* indicates whether the client-generated dependency tree is from Chromium
```
kazuho: Could we add "is_" prefix to this boolean attribute? I think we tend to do that.

hfujita: Sure. Will do.

hfujita:
@kazuho For some reason I can't see what https://github.com/h2o/h2o/pull/2217/files#r368538998 refers to. Just wanted to make sure your only remaining concern is about adding `is_` to `chromium_dependency_tree`? Did you want to point out anything else?

kazuho:
@hfujita Excuse me, I was referring to an inline comment that was marked resolved. However, it seems that links to those do not work, unless you open the resolved comments.

I've unresolved the corresponding one, so I hope you can now track the link. FWIW, it's the discussion about not sending the settings parameter to the client.

lib/http2/frame.c
hfujita: Turns out we both misunderstood this code snippet.

This is actually about *accepting* `SETTINGS_DEPRECATE_HTTP2_PRIORITIES` from a client. At this moment this doesn't do any meaningful work other than checking the value range. See the patch `[http2] Accept SETTINGS_DEPRECATE_HTTP2_PRIORITIES` for details.

FYI, the patch to *send* `SETTINGS_DEPRECATE_HTTP2_PRIORITIES` has already been dropped from this PR, so it is not sent to a client. If you prefer, I can also drop this patch (`[http2] Accept SETTINGS_DEPRECATE_HTTP2_PRIORITIES`).

hfujita:
@kazuho Ah I see. Turns out it is not about sending the settings parameter, instead it is about *accepting* the parameter. Made a [comment](https://github.com/h2o/h2o/pull/2217#discussion_r372103601) there.

lib/http2/frame.c
kazuho: Ah. Thank you for the explanation.

I think it might be a good idea to drop all code related to SETTINGS_DEPRECATE_HTTP2_PRIORITIES, even though the code looks fine and I appreciate your efforts.

The reasons are:
* we do not use the flag in any ways
* the specification might change
* we might use the value differently. For example, rather than retaining a boolean variable, we might use the existence (or lack of) the settings parameter to initiate different scheduler.

We can resurrect the code when necessary.

hfujita: 👍 Totally makes sense. I'll drop this patch. Thanks!

include/h2o/http2_internal.h
hfujita: Done.

lib/http2/frame.c
hfujita: Done.

hfujita:
@kazuho I believe I've addressed all of your comments, and some commits have been squashed as I said earlier. This should be good to go once the CI comes back with green.

hfujita:
@kazuho Thank you a lot for your inputs to improve this PR! 😄

> Re rebasing - generally speaking I'd appreciate it if you could just add commits (or merges) once the PR has gone review. That way, it is easier to find out what exactly has changed since the last review, also does not break the local repositories of others.

I understand -- will do so from the next time.
After approval, do you still prefer to keep every individual fixup patch unsquashed?

Also, I just noticed one test failure in `fuzz` configuration. Is this something I should worry about?

kazuho:
> After approval, do you still prefer to keep every individual fixup patch unsquashed?

Yes. I'd argue that a branch should not be rebased (i.e. force-pushed) once it becomes a shared property. I think it is clear that the line has been crossed when the PR goes through the first review. But that could be earlier than that.

H2O is a small project, we do not need to limit the number of commits. Besides, in case the PR becomes really messy, we always have the choice of squash-merging the PR.

> Also, I just noticed one test failure in fuzz configuration. Is this something I should worry about?

Please look at the log (maybe you need to look at the raw log), and see why it's failing. If it looks like a non-issue (e.g., port already used by another process), please feel free to ask me, or push an empty commit, for rerunning the CI.

In this particular case, I think the CI *has not been run* on the latest commit (maybe due to forced push).

I assume that you would be pushing a small fix soon, that would invoke a new CI process, and hopefully everything would go green.

hfujita:
> Yes. I'd argue that a branch should not be rebased (i.e. force-pushed) once it becomes a shared property.

I see. So you treat a brach associated with a PR is a shared property, even if it sits in my personal fork of h2o.

> I assume that you would be pushing a small fix soon, that would invoke a new CI process, and hopefully everything would go green.

Yep. I'll push a few more commits (no rebase this time, I promise) and see the results.

Regarding CI jobs, would you mind explaining the difference between `travis-ci/pr` and `travis-ci/push`?

kazuho:
> Regarding CI jobs, would you mind explaining the difference between travis-ci/pr and travis-ci/push?

Maybe https://mahdi.blog/travis-ci-pr-push/ is a good pointer?

hfujita:
Ah! I thought they were project-specific job configurations. Thanks!

PR #2259, opened 2020-02-11
http3: Rate limit the sendmsg failed error

deweerdt:
In practice, these errors tend to happen all at the same time

### DIFF
--- a/lib/http3/common.c
+++ b/lib/http3/common.c
@@ -57,6 +57,51 @@ struct st_h2o_http3_ingress_unistream_t {
 
 const ptls_iovec_t h2o_http3_alpn[1] = {{(void *)H2O_STRLIT("h3-25")}};
 
+static struct {
+    pthread_mutex_t lock;
+    time_t last_reported;
+    size_t failures;
+    size_t successes;
+    int last_errno;
+} track_sendmsg = {
+    .lock = PTHREAD_MUTEX_INITIALIZER,
+};
+
+static void track_sendmsg_failure(void)
+{
+    __sync_fetch_and_add(&track_sendmsg.failures, 1);
+    track_sendmsg.last_errno = errno;
+}
+
+static void track_sendmsg_success(void)
+{
+    __sync_fetch_and_add(&track_sendmsg.successes, 1);
+}
+
+static void track_sendmsg_flush(h2o_loop_t *loop)
+{
+    uint64_t now = h2o_now(loop);
+    if (track_sendmsg.last_reported + 60000 < now) {
+        /* this is racy, and might under report successes when there is a
+           failure, but allows to avoid taking the lock in the happy path */
+        if (track_sendmsg.failures == 0) {
+            track_sendmsg.last_reported = now;
+            track_sendmsg.successes = 0;
+            return;
+        }
+        pthread_mutex_lock(&track_sendmsg.lock);
+        if (track_sendmsg.last_reported + 60000 < now) {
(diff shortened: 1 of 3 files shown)


deweerdt:
Thanks for your feedback, @kazuho . What do you think of ca8a390?

lib/http3/common.c
```diff
@@ -57,6 +57,51 @@ struct st_h2o_http3_ingress_unistream_t {
 
 const ptls_iovec_t h2o_http3_alpn[1] = {{(void *)H2O_STRLIT("h3-25")}};
 
+static struct {
+    pthread_mutex_t lock;
+    time_t last_reported;
+    size_t failures;
+    size_t successes;
```
kazuho: I think that the type of counters should be `uint64_t`. These counts are unrelated to the amount of memory that can be allocated, and needs to be greater than 32-bit unsigned int even in 32-bit environments.

lib/http3/common.c
```diff
@@ -57,6 +57,51 @@ struct st_h2o_http3_ingress_unistream_t {
...
+
+static void track_sendmsg_flush(h2o_loop_t *loop)
+{
+    uint64_t now = h2o_now(loop);
+    if (track_sendmsg.last_reported + 60000 < now) {
+        /* this is racy, and might under report successes when there is a
+           failure, but allows to avoid taking the lock in the happy path */
+        if (track_sendmsg.failures == 0) {
+            track_sendmsg.last_reported = now;
+            track_sendmsg.successes = 0;
```
kazuho: I think it might be a good idea to define `successes` and `failures` as counters from the start of the process, rather than counters from the previous log emission. That avoids the problem of resetting the values.

When it becomes time to log, we can simply copy the values, compare them with the previous copy, and log the difference.

lib/http3/common.c
```diff
@@ -134,8 +179,11 @@ int h2o_http3_send_datagram(h2o_http3_ctx_t *ctx, quicly_datagram_t *p)
 
         /* Temporary failure to send a packet is not a permanent error fo the connection. (TODO do we want do something more
          * specific?) */
-        perror("sendmsg failed");
+        track_sendmsg_failure();
+    } else {
+        track_sendmsg_success();
     }
+    track_sendmsg_flush(ctx->loop);
```
kazuho: I wonder if we need to log anything when there is no loss. To paraphrase, it might make sense to convert `track_sendmsg_flush` into a subroutine of `track_sendmsg_failure`.

The other benefit of merging `track_sendmsg_flush` into `track_sendmsg_failure` is that we can get rid of `last_errno`. `errno` would always be available when we log the message.

deweerdt: If we do so, we might wait for a long time for messages to appear, would you be ok with that?

kazuho: How about emitting a message when both of the following conditions are met:
* a new failure event occurs
* at least 60 seconds have elapsed since the last emission of the message

That would guarantee that we'd see the message at least once a minute, when loss is happening. I think that would be fine?

kazuho:
Thank you for the changes. I've filed https://github.com/deweerdt/h2o/pull/3, please let me know what you think.

deweerdt:
That looks good, thank you! I'm going to give it a test today

deweerdt:
Tested and merged, thank you @kazuho . Here's a sample log:
```
Feb 13 20:46:52 localhost h2o[192687]: sendmsg failed 73 times, succeeded: 115878 times, over the last minute: Resource temporarily unavailable                               
```

kazuho:
Good to know that the changes look okay and works.

PR #2319, opened 2020-05-11
[doc] explain `-DWITH_DTRACE` and `-DWITH_H2OLOG` in cmake options

gfx:
It is not clear to enable DTrace support on Linux, so this PR explains how to enable DTrace support both on Linux and macOS.


srcdoc/install.mt
```diff
@@ -78,6 +78,11 @@ This option specifies the directory to which H2O will be installed (default: <co
 This option instructs whether or not to build the standalone server with support for <a href="configure/mruby.html">scripting using mruby</a>.
 It is turned on by default if the prerequisites (<a href="https://www.gnu.org/software/bison/">bison</a>, <a href="https://www.ruby-lang.org/">ruby</a> and the development files<?= $ctx->{note}->(q{<code>mkmf</code> - a program for building ruby extensions is required.  In many distributions, the program is packaged as part of <code>ruby-dev<code> or <code>ruby-devel</code> package.}) ?>) are found.
 </dl>
+<dt><code>-DWITH_DTRACE=<i>on</i>|<i>off</i></code></dt>
+<dd>
+This option instructs whether or not to enable DTrace support.
+It is turned on by default if the prerequisites (<a href="https://sourceware.org/systemtap/">SystemTap</a> on Linux, or <a href="https://github.com/h2o/h2o/wiki/macOS">macOS without System Integirty Protection</a>) are found.
```
kazuho, line 84: The description for macOS is not technically correct. The OS having SIP depends on if you can use dtrace; not if you can build h2o (or how we build h2o).

gfx:
I'd like to push it forward, but I'd like to change the description.

The `-DWITH_DTRACE` (and `-DWITH_H2OLOG`) itself doesn't matter. Probably nobody needs to use the option(s) directly. Instead,  I'd like to explain the feature and dependencies.

gfx:
How about this?

Screen Shot 2020-07-08 at 11 59 34

PR #2320, opened 2020-05-11
Add the ability to specify address family for proxy connections

csjperon:
When the family key is used, the address family for the backend specification
can be controlled. Legal values for are inet and inet6. If nothing is specified,
an address family of PF_UNSPEC is used, which will result in both IP and
IPv6 (if present) being used.

Example:

proxy.reverse.url:
  - url: http://some.host.com:8080
    family: inet

### DIFF
--- a/doc/configure/proxy_directives.html
+++ b/doc/configure/proxy_directives.html
@@ -168,6 +168,20 @@ Forwards the requests to the specified backends, and proxies the response.
 </code></pre>
 </div>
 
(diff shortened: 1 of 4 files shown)


csjperon:
This should address the issue described in https://github.com/h2o/h2o/issues/1285 as well

csjperon:
Fix a knit that was picked up by the fuzzer

PR #2324, opened 2020-05-14
remove unused variable

i110:


syohex:
This is duplicated with #2318

PR #2357, opened 2020-06-16
Add workaround for CIFuzz build bug [WIP]

jfoote:
CIFuzz currently does not support fuzzing of forks. See https://github.com/google/oss-fuzz/issues/3731. This changes the configuration so that CIFuzz does not run on forks, i.e. does not run on PRs from forked repos. I believe https://github.com/h2o/h2o/pull/2356 (presently) shows an example failure.

Note that I copied this approach from [systemd's CIFuzz configuration](https://github.com/systemd/systemd/blob/master/.github/workflows/cifuzz.yml).

This change should be rolled back when the CIFuzz bug is fixed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,21 +3,23 @@ on: [pull_request]
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
-    steps:
-    - name: Build Fuzzers
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
-      with:
-        oss-fuzz-project-name: 'h2o'
-        dry-run: false
-    - name: Run Fuzzers
-      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
-      with:
-        oss-fuzz-project-name: 'h2o'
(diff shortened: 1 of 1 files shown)


jfoote:
https://github.com/h2o/h2o/pull/2353 appears to show a related failure (presently)

jfoote:
I added [WIP] to the title while I debug this

jfoote:
I believe we are running into a separate issue here. The oss-fuzz documentation states that listing supported sanitizers in a project's yaml [is optional](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers). It appears that CIFuzz requires supported sanitizers to be included in this file ([ref](https://github.com/google/oss-fuzz/blob/26e8d7c7728096edf55a1fb6d0ecbc4b2dae6afa/infra/cifuzz/cifuzz.py#L508-L529)). I believe this causes the CIFuzz Github Action to fail for projects that have not listed support for address sanitizer explicitly in their respective project YAML. We can try to fix this via a PR to oss-fuzz.

jfoote:
PR is here: https://github.com/google/oss-fuzz/pull/3995

PR #2362, opened 2020-06-29
Fix h2olog to follow the latest quicly-probes.d (st_quicly_stats_t)

gfx:
Reported by @syohex  at https://github.com/h2o/h2o/pull/2353#issuecomment-650887319

@kazuho `st_quicly_stats_t` is empty and thus `conn_stats` event is useless, but if you need some fields, I'll add them.


syohex:
Thanks, I have confirmed that this patch fixes my build issue.

PR #2397, opened 2020-07-30
refactor proceed_handshake

i110:
Thanks to https://github.com/h2o/h2o/commit/a42769ce4dceee6c57e9453e59af5e67ab9d267f which made picotls no longer optional, now we can have a chance to refactor complicated `proceed_handshake`.


kazuho (approved):
My jaw dropped, literally. I am fascinated by how easy to read the code is now! It has been a mess, I did not even expect that it can become as clean as this.

I've made editorial adjustments, but this PR is good to go. 💯💯💯💯💯💯💯

lib/common/socket.c
```diff
@@ -1209,6 +1175,87 @@ static void proceed_handshake(h2o_socket_t *sock, const char *err)
     on_handshake_complete(sock, err);
 }
 
+/**
+ * Called when it is still untertain which of the two TLS stacks (picotls or OpenSSL) should handle the handshake.
```
kamipo, line 1179: Maybe "uncertain"?

```suggestion
 * Called when it is still uncertain which of the two TLS stacks (picotls or OpenSSL) should handle the handshake.
```

PR #2428, opened 2020-08-23
consume buffer at once

i110:
This PR fixes a coverity defect https://scan7.coverity.com/reports.htm#v29627/p11912/fileInstanceId=85147268&defectInstanceId=16919126&mergedDefectId=1496212.

As of https://github.com/h2o/h2o/pull/2010 `handle_one_body_fragment` function may invoke `cleanup_connection`, so we have to consume the buffer before the connection gets freed (or reused for the next request).


hfujita (approved):
Thank you for the patch! @i110 

@kazuho do you think we can merge this?

PR #2438, opened 2020-08-31
Transparent Reverse Proxy

cwyang:
Hi, I send several patches for Transparent Reverse Proxy implementation of issue https://github.com/h2o/h2o/issues/2416

Following three patches are simple APIs:
* add h2o_socket_setport()
* h2o_cache_t: H2O_CACHE_FLAG_AGE_UPDATE …
* h2o_socket_connect_tproxy(.. srcaddr, srclen, dstaddr, dstlen ..) …

I use simple stdout h2o_pinfo() macro during development session:
* h2o/util.h …

This patch adds configuration and enables IP_TRANSPARENT on listen socket
* tproxy: listen.transparent configuration …

These two patches are main contribution.
First patch lets us allocate connection pool and socket pool. We maintain connection pool and socketpool per 5-tuples.
Second patch uses the former patch to implement src-address spoofed upstream session. The patch also
provides TPROXY socket pool target, which has`proxy.reverse.url="http[s]://tproxy/` configuration:
* connection pool and socket pool can be allocated …
* tproxy: Transparent Reverse Proxy …

For test, I separate client namespace and h2o namespace and setup routing rule and iptables rule to feed the packet to h2o.
Web browser repeatedly send web request and the request was served from h2o transparently to both ends.
I wonder whether I can make automatic testing for this.

Best regards,
Chul-Woong

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,7 @@ SET(LIB_SOURCE_FILES
     lib/handler/headers.c
     lib/handler/mimemap.c
     lib/handler/proxy.c
+    lib/handler/proxy/tproxy.c
     lib/handler/redirect.c
     lib/handler/reproxy.c
     lib/handler/throttle_resp.c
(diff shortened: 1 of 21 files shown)


cwyang:
I'll sleep on this PR and try to make more general PR.

Take care,
Chul-Woong

PR #2453, opened 2020-09-15
h2olog: drop the root privilege once BPF is set up

toru:
Technically `h2olog` only needs root privilege to:

1. Attach the BPF tracepoints
2. Open the BPF ring buffer

Therefore drop the root privilege as soon as the two operations are complete. Error handling may seem unforgiving but it follows a consistent pattern in `h2olog`.

### DIFF
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -43,6 +43,7 @@ Usage: h2olog -p PID
 Other options:
     -h Print this help and exit
     -d Print debugging information (-dd shows more)
+    -r Run without dropping root privilege
     -w Path to write the output (default: stdout)
 )",
            H2O_VERSION);
@@ -234,7 +235,7 @@ int main(int argc, char **argv)
     std::vector<std::string> response_header_filters;
     int c;
     pid_t h2o_pid = -1;
-    while ((c = getopt(argc, argv, "hdp:t:s:w:")) != -1) {
+    while ((c = getopt(argc, argv, "hdrp:t:s:w:")) != -1) {
         switch (c) {
         case 'p':
             h2o_pid = atoi(optarg);
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -133,7 +133,12 @@ static void drop_root_privilege(void)
             fprintf(stderr, "Error: failed to read the SUDO_GID env variable\n");
             exit(EXIT_FAILURE);
         }
-        gid_t gid = atoi(sudo_gid);
+        errno = 0;
+        gid_t gid = (gid_t)strtol(sudo_gid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: overflow while parsing SUDO_GID\n");
+            exit(EXIT_FAILURE);
+        }
         if (gid != 0 && setgid(gid) != 0) {
             fprintf(stderr, "Error: failed to drop the root group\n");
             exit(EXIT_FAILURE);
@@ -145,8 +150,13 @@ static void drop_root_privilege(void)
             fprintf(stderr, "Error: failed to read the SUDO_UID env variable\n");
             exit(EXIT_FAILURE);
         }
-        uid_t uid = atoi(sudo_uid);
-        if (uid > 0 && setuid(uid) != 0) {
+        errno = 0;
+        uid_t uid = (uid_t)strtol(sudo_uid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: overflow while parsing SUDO_UID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (uid != 0 && setuid(uid) != 0) {
             fprintf(stderr, "Error: failed to drop the root user\n");
             exit(EXIT_FAILURE);
         }
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -138,7 +138,7 @@ static void drop_root_privilege(void)
             errno = 0;
             gid_t gid = (gid_t)strtol(sudo_gid, NULL, 10);
             if (errno != 0) {
-                fprintf(stderr, "Error: overflow while parsing SUDO_GID\n");
+                fprintf(stderr, "Error: failed to parse SUDO_GID\n");
                 exit(EXIT_FAILURE);
             }
             if (gid != 0 && setgid(gid) != 0) {
@@ -154,7 +154,7 @@ static void drop_root_privilege(void)
         errno = 0;
         uid_t uid = (uid_t)strtol(sudo_uid, NULL, 10);
         if (errno != 0) {
-            fprintf(stderr, "Error: overflow while parsing SUDO_UID\n");
+            fprintf(stderr, "Error: failed to parse SUDO_UID\n");
             exit(EXIT_FAILURE);
         }
         if (uid != 0 && setuid(uid) != 0) {
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -141,7 +141,7 @@ static void drop_root_privilege(void)
             exit(EXIT_FAILURE);
         }
         if (setgid(gid) != 0) {
-            perror("Error: setgid(2) failed\n");
+            perror("Error: setgid(2) failed");
             exit(EXIT_FAILURE);
         }
         const char *sudo_uid = getenv("SUDO_UID");
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -131,7 +131,7 @@ static void drop_root_privilege(void)
     if (getuid() == 0) {
         const char *sudo_gid = getenv("SUDO_GID");
         if (sudo_gid == NULL) {
-            fprintf(stderr, "Error: failed to read the SUDO_GID env variable\n");
+            fprintf(stderr, "Error: the SUDO_GID environment variable is not set\n");
             exit(EXIT_FAILURE);
         }
         errno = 0;
@@ -146,7 +146,7 @@ static void drop_root_privilege(void)
         }
         const char *sudo_uid = getenv("SUDO_UID");
         if (sudo_uid == NULL) {
-            fprintf(stderr, "Error: failed to read the SUDO_UID env variable\n");
+            fprintf(stderr, "Error: the SUDO_UID environment variable is not set\n");
             exit(EXIT_FAILURE);
         }
         errno = 0;
--- a/src/h2olog/h2olog.cc
+++ b/src/h2olog/h2olog.cc
@@ -228,6 +228,7 @@ int main(int argc, char **argv)
     }
 
     int debug = 0;
+    int drop_root = 1;
     FILE *outfp = stdout;
     std::vector<std::string> event_type_filters;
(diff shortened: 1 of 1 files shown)


src/h2olog/h2olog.cc
```diff
@@ -125,6 +125,34 @@ static void show_process(pid_t pid)
...
+        }
+    }
+    if (getuid() == 0) {
+        const char *sudo_uid = getenv("SUDO_UID");
+        if (sudo_uid == NULL) {
+            fprintf(stderr, "Error: failed to read the SUDO_UID env variable\n");
+            exit(EXIT_FAILURE);
+        }
+        uid_t uid = atoi(sudo_uid);
+        if (uid > 0 && setuid(uid) != 0) {
```
gfx, line 149:
```suggestion
        if (uid != 0 && setuid(uid) != 0) {
```

kazuho (commented):
Thank you for the pull request. I think this is a nice change; left some thoughts. PTAL.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +125,44 @@ static void show_process(pid_t pid)
     infof("Attaching pid=%d (%s)", pid, cmdline);
 }
 
+static void drop_root_privilege(void)
+{
+    if (getgid() == 0) {
```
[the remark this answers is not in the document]
gfx: In addition, h2olog already checks euid at the startup: https://github.com/h2o/h2o/blob/34e41a2013dbdfaed776fcc71e64962b1b207d9a/src/h2olog/h2olog.cc#L239-L242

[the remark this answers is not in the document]
toru: Thanks you two. The `setgid()` code is now nested under `if (getuid() == 0)` - e85fc361

src/h2olog/h2olog.cc
toru: Thank you @gfx. This is addressed in: df7ba942

src/h2olog/h2olog.cc
```diff
@@ -125,6 +125,44 @@ static void show_process(pid_t pid)
...
+        const char *sudo_gid = getenv("SUDO_GID");
+        if (sudo_gid == NULL) {
+            fprintf(stderr, "Error: failed to read the SUDO_GID env variable\n");
+            exit(EXIT_FAILURE);
```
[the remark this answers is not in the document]
toru: Thank you for pointing out my blind spot. Yes, adding a command line option to skip privdrop makes sense to me. This is addressed in: ce5e7b6 and 35ce6ca

src/h2olog/h2olog.cc
```diff
@@ -190,12 +229,13 @@ int main(int argc, char **argv)
     }
 
     int debug = 0;
+    int drop_root = 1;
```
kazuho, line 232: I think it might be less confusing if we rename this to `preserve_root` or something to avoid using inverted definitions against the command line option.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,44 @@ static void show_process(pid_t pid)
     infof("Attaching pid=%d (%s)", pid, cmdline);
 }
 
+static void drop_root_privilege(void)
+{
+    if (getuid() == 0) {
+        if (getgid() == 0) {
```
kazuho, line 132: I think it might be possible remove this `if`, is it necessary? I think that the behavior we want is "if running as root, drop root privileges."

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,44 @@ static void show_process(pid_t pid)
...
+            const char *sudo_gid = getenv("SUDO_GID");
+            if (sudo_gid == NULL) {
+                fprintf(stderr, "Error: failed to read the SUDO_GID env variable\n");
+                exit(EXIT_FAILURE);
+            }
+            errno = 0;
+            gid_t gid = (gid_t)strtol(sudo_gid, NULL, 10);
+            if (errno != 0) {
+                fprintf(stderr, "Error: overflow while parsing SUDO_GID\n");
```
kazuho, line 141:
```suggestion
                fprintf(stderr, "Error: failed to parse SUDO_GID\n");
```
... or something alike. I think that the parsing might fail for reasons other than overflow?

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,44 @@ static void show_process(pid_t pid)
...
+                exit(EXIT_FAILURE);
+            }
+            if (gid != 0 && setgid(gid) != 0) {
```
kazuho, line 144: I think we'd either remove `gid != 0` or call it an error.

Though it depends on how we define what "dropping root privileges" mean. If we are to define the act as switching to SUDO_UID / SUDO_GID, then simply removing this condition is fine. If we are to define the act as "dropping root privileges" in a strict manner, then we should call it an error.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,44 @@ static void show_process(pid_t pid)
...
+        }
+        const char *sudo_uid = getenv("SUDO_UID");
+        if (sudo_uid == NULL) {
+            fprintf(stderr, "Error: failed to read the SUDO_UID env variable\n");
+            exit(EXIT_FAILURE);
+        }
+        errno = 0;
+        uid_t uid = (uid_t)strtol(sudo_uid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: overflow while parsing SUDO_UID\n");
```
kazuho, line 157: Please see my comment above.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,44 @@ static void show_process(pid_t pid)
...
+            exit(EXIT_FAILURE);
+        }
+        if (uid != 0 && setuid(uid) != 0) {
```
kazuho, line 160: Here also.

kazuho (commented):
Thank you for the changes. Left some comments.

src/h2olog/h2olog.cc
toru: Nice! yes, thank you for pointing this out. This definitely requires few brain cycles to grok.

src/h2olog/h2olog.cc
toru: Agreed that we need to decide on the definition, and I'm good with simply: "switching to SUDO_UID / SUDO_GID" as you mentioned. I see your point about needing to be more unforgiving if we had chosen a stricter definition. Thanks!

toru:
@kazuho As always, thank you for the thorough review. I've addressed your feedback in the following commits:

108bdbb - rename the drop_root option var to preserve_root
2e9fa21 - update the error message
7937ab9 -  call setuid() and setgid() regardless of the gid/uid

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
...
+            exit(EXIT_FAILURE);
+        }
+        errno = 0;
+        gid_t gid = (gid_t)strtol(sudo_gid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: failed to parse SUDO_GID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (setgid(gid) != 0) {
+            fprintf(stderr, "Error: failed to drop the root group\n");
```
kazuho, line 144: I think we might want to use `perror` for syscalls, because the cause of the failure could be anything.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
...
+            exit(EXIT_FAILURE);
+        }
+        errno = 0;
+        uid_t uid = (uid_t)strtol(sudo_uid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: failed to parse SUDO_UID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (setuid(uid) != 0) {
+            fprintf(stderr, "Error: failed to drop the root user\n");
```
kazuho, line 159: Same here.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
     infof("Attaching pid=%d (%s)", pid, cmdline);
 }
 
+static void drop_root_privilege(void)
+{
+    if (getuid() == 0) {
+        const char *sudo_gid = getenv("SUDO_GID");
+        if (sudo_gid == NULL) {
+            fprintf(stderr, "Error: failed to read the SUDO_GID env variable\n");
```
kazuho, line 134: Should we better say that the environment variable is not set, rather than "failed to read"?

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
...
+            fprintf(stderr, "Error: failed to parse SUDO_GID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (setgid(gid) != 0) {
+            fprintf(stderr, "Error: failed to drop the root group\n");
+            exit(EXIT_FAILURE);
+        }
+        const char *sudo_uid = getenv("SUDO_UID");
+        if (sudo_uid == NULL) {
+            fprintf(stderr, "Error: failed to read the SUDO_UID env variable\n");
```
kazuho, line 149: Same here.

kazuho (commented):
Thank you for the changes. LGTM (with minor suggestions).

src/h2olog/h2olog.cc
toru: Thank you for noticing. This did not occur to me at the time. Addressed in: e2aef5e

src/h2olog/h2olog.cc
toru: Great catch! Addressed in: b6c5d3f

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
...
+            exit(EXIT_FAILURE);
+        }
+        errno = 0;
+        gid_t gid = (gid_t)strtol(sudo_gid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: failed to parse SUDO_GID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (setgid(gid) != 0) {
+            perror("Error: setgid(2) failed\n");
```
kazuho, line 144:
```suggestion
            perror("Error: setgid(2) failed");
```

perror will append a colon, the error message, and a newline.

src/h2olog/h2olog.cc
```diff
@@ -125,6 +126,42 @@ static void show_process(pid_t pid)
...
+            exit(EXIT_FAILURE);
+        }
+        errno = 0;
+        uid_t uid = (uid_t)strtol(sudo_uid, NULL, 10);
+        if (errno != 0) {
+            fprintf(stderr, "Error: failed to parse SUDO_UID\n");
+            exit(EXIT_FAILURE);
+        }
+        if (setuid(uid) != 0) {
+            perror("Error: setuid(2) failed\n");
```
kazuho, line 159:
```suggestion
            perror("Error: setuid(2) failed");
```

kazuho (approved):

PR #2477, opened 2020-10-31
http3: Link connection to list after accept succeeds

hfujita:
This is a follow up/partial revert of https://github.com/h2o/h2o/pull/2451
(48bd91993b53549473bde7f375a28169e7ebe04e) to prevent NULL pointer
dereference.

When `quicly_accept` fails, `conn->h3.super.quic` is left unset
(NULL).  And if WITH_DTRACE is set, `on_h3_destroy` invokes a probe,
which then invokes `H2O_CONN_IS_PROBED`, which tries to access
`conn->h3.super.quic`. (`skip_tracing` is set to `get_skip_tracing`.)

This patch solves this crash and use-after-free solved in https://github.com/h2o/h2o/pull/2451
by inserting a connection to the connection list only when
`quicly_accept` succeeds.

### DIFF
--- a/lib/http3/server.c
+++ b/lib/http3/server.c
@@ -1553,7 +1553,6 @@ h2o_http3_conn_t *h2o_http3_server_accept(h2o_http3_server_ctx_t *ctx, quicly_ad
     conn->scheduler.uni.active = 0;
     conn->scheduler.uni.conn_blocked = 0;
     memset(&conn->_conns, 0, sizeof(conn->_conns));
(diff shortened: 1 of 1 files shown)


kazuho (approved):
Thank you for the PR. Looks good to me.

We should not be invoking H3S_DESTROY when `quicly_accept` fails, because we do not invoke `H3_ACCEPT` in that case.

hfujita:
Thanks, merged!

PR #2482, opened 2020-11-07
h2o-httpclient: handle -H "host: ..." arg as the authority as what http1.c does

gfx:
Motivation: https://github.com/h2o/h2o/pull/2481#issuecomment-722921807


kazuho (changes requested):
As discussed elsewhere, httpclient is HTTP-version agnostic. That means that it should not be dealing with the host header field, as that header field is specific to HTTP/1.

If there is a place that uses the host header field where the *only* source of the request URI should be the URI being supplied, then that would be the place to fix.

gfx:
Thanks for your comments. Will reconsider it in another way.

PR #2503, opened 2020-12-01
extract https://github.com/h2o/quicly @ 8648403 () at deps/quicly

hfujita:
This is to test h2o with newer quicly, namely with h2o/quicly#421.


kazuho (approved):
LGTM. Please feel free to merge now (or merge it as part of a different PR uses this change).

PR #2504, opened 2020-12-02
cmake: state libmruby.a and libonigmo.a are generated on building mruby

gfx:
Some cmake generators like Ninja cannot handle implicitly generated files, failing with messages like (in case of Ninja):

> ninja: error: 'mruby/host/lib/libmruby.a', needed by 'h2o', missing and no known rule to make it

The solution is just to define files with `BYPRODUCTS libmruby.a libonigumo.a` in `ADD_CUSTOM_TARGET`.

With this patch, `cmake -G Ninja -Bb . && cmake --build b` works!


kazuho (approved):
LGTM.

PR #2527, opened 2020-12-25
retrieve "fuzz" test

gfx:
amending https://github.com/h2o/h2o/pull/2518 which accidentally deleted this test


kazuho (approved):
Nice catch!

PR #2541, opened 2021-01-12
ci: update iovisor/bcc from a revision of 0.11.0-6021958a to 0.12.0

gfx:
I've made sure an h2olog E2E test (https://github.com/h2o/h2o/pull/2520) works with this image. If it looks good, I'll push the image to DockerHub, restart this CI, and then merge this PR.

### DIFF
--- a/misc/docker-ci/Dockerfile.ubuntu2004
+++ b/misc/docker-ci/Dockerfile.ubuntu2004
@@ -64,21 +64,13 @@ RUN sudo cpanm --notest https://github.com/tokuhirom/Test-TCP.git@2.21
 RUN curl -Ls https://github.com/i110/h2spec/releases/download/v2.2.0-4e8cc7e/h2spec_linux_amd64.tar.gz | tar zx -C /usr/local/bin
 
 # bcc and bpftrace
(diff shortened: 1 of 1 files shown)


kazuho (approved):
Fantastic! Thank you for all the efforts, I understand that this has been a complicated task. Please proceed.

gfx:
Thanks. Proceeding the new one in the branch https://github.com/h2o/h2o/tree/gfx/h2olog-e2e-tests-canary , and I have found the tests stick in Docker containers. Fortunately, I can reproduce this problem in a local Docker.

gfx:
Confirmed that h2olog testing works with this image! (https://github.com/h2o/h2o/runs/1745751370).

The points are:

* BCC needs the `linux-headers` package that corresponds to the runtime kernel
* Installing linux-headers on `docker build` (e.g. 14f2950) on my local VM does not work because it does not match the runtime kernel
* The only way to provide the right linux-headers is to share host's /usr/src with containers, i.e. adding `-v /usr/src:/usr/src:ro` to `docker run` options

I'm working on h2olog E2E testing on CI at https://github.com/h2o/h2o/pull/2520 to apply the above points.

PR #2547, opened 2021-01-26
[h2olog] Fix parameter position in example usage

hfujita:
`-p` should appear right in front of a process ID.


gfx (approved):
Good catch! Thanks.

PR #2549, opened 2021-01-29
fix deprecated warnings by s/TLSv1_client_method/SSLv23_client_method/

gfx:
`TLSv1_client_method` has been deprecated in OpenSSL 1.1.0. `TLS_client_method` is not available in older OpenSSL, but `SSLv23_client_method` is portable and not deprecated, which is `#define SSLv23_client_method    TLS_client_method` as of OpenSSL 1.1.1i.


kazuho (approved):
Thank you for the fix. As you correctly point out, `SSLv23_*` are the stable APIs that we should be using.

PR #2561, opened 2021-02-10
Make sure the cid is initialized, even in the case we the keyset is not found

deweerdt:


hfujita (commented):
@deweerdt Change looks good but could you give me some more context on in what case this would be needed?

deweerdt:
@hfujita without the change the reader in `process_packets` reads uninitialized memory here https://github.com/h2o/h2o/blob/master/lib/http3/common.c#L540-L541

It doesn't materially affect the outcome of the processing: random memory is going to behave in the same way as random cids set by the client.

kazuho (commented):
Thank you for the PR.

First of all, regarding the value that we should set when decryption fails, should it be {.node_id = max, .thread_id = max} rather than {.node_id = 0, .thread_id = max}?

I did suggest using the latter when we discussed offline, but now I tend to think that it'd be better to also set node_id to an unreasonably large value to minimize the possibility of broken packets being forwarded across the nodes when people use a node map with a zero-based index.

Regarding where the change should be applied, I wonder if it makes sense to change [quicly_decode_packet function of lib/quicly.c](https://github.com/h2o/quicly/blob/e6df839/lib/quicly.c#L614-L619) instead, so that `packet->cid.dest.plaintext` would be set to the special value if `cid_encryptor->decrypt_cid` returns SIZE_MAX or if `cid_encryptor` is NULL?

This PR is based on the assumption that the contract of `decrypt_cid` is to return the special value if it fails. I'm not sure if we'd want to have such a contract when the callback API has the capability of returning an error. We could simply rely on the caller to detect and set certain value when the callback returns an error.

Also, that fallback value should be used when a CID encryptor is not registered - we need to change that code path in `quicly_decode_packet` anyways.

deweerdt:
Closing this in favor of the quicly PR.

PR #2624, opened 2021-03-17
amends b27ba5a of #2602 so that the disposal of `h2o_req_t` would be delayed only when the caller is the generator

deweerdt:


kazuho (changes requested):
Adding a note that this issue has been fixed differently in master as #2625 and that kazuho/connect-udp should be fixed by merge master.

PR #2628, opened 2021-03-20
Allow a non-first host entry to be marked as the default one to use when none match.

jhatala:
This allows specific entries to be listed first before a wildcard entry and still to route requests without host (or :authority) to the wildcard entry rather than to the first specific entry.

Example:

```
hosts:
  "www.example.com":
    paths:
      /:
        file.dir: examples/doc_root1
  "*.example.com":
    paths:
      /:
        file.dir: examples/doc_root2
  "*":
    default: ON
    paths:
      /:
        file.dir: examples/doc_root3
```

In this example, requests without a Host header will be handled by the third host entry.

Without a way to mark an entry as the default entry, requests without a Host header can only be handled by the first host entry in the configuration.


jhatala:
Replaced by #2632.

PR #2639, opened 2021-03-25
Add error details in a `Proxy-Status` header when a CONNECT request fails.

jhatala:
This is an implementation of [draft-ietf-httpbis-proxy-status-03](https://tools.ietf.org/html/draft-ietf-httpbis-proxy-status-03) that allows the client to distinguish the different ways in which CONNECT requests can fail.

Example:
```
proxy-status: h2o; error=dns_error; rcode=NXDOMAIN; details="hostname does not exist"
```

This is only enabled when `proxy.connect.proxy-status: ON` is applied.

`proxy-status.identity` at the global level of the configuration configures what will be the token or string at the start of the `proxy-status` header.  It defaults to "h2o".


jhatala:
Subsumed by #2640.

PR #2647, opened 2021-03-28
In the CONNECT handler: close the attempted socket on timeout or failure.  [kazuho/connect-udp]

jhatala:
This is PR is the same as #2645 but for `kazuho/connect-udp`.  The active ingredient is c01f013.

----

This fixes two problems in the handling of the `CONNECT` method:
1. If a hostname resolves to multiple IPs and the `connect` to the first one or two on the list fails asynchronously (rather than immediately in the `connect` call), then we leak the file-descriptor(s) for the failed socket(s).
2. If the `connect` fails asynchronously after `proxy.timeout.connect` has already passed and returned an error, then we will attempt to write another error response when the asynchronous event happens.  That attempt is fatal in H3 at least.

c57deed (reverted in dff7f5f) contains a repro case for problem number 2.  The asynchronous failure in this case is `ETIMEDOUT`, which arrives seconds or minutes after `connect` returned `EAGAIN`, depending on the kernel configuration.  I have had good results in reproducing the bug after setting `sysctl net.ipv4.tcp_syn_retries=2`.  The test in c57deed however is not suitable to become a permanent regression test, that is why dff7f5f reverts it.


jhatala:
`kazuho/connect-udp` has received #2645 via branch `master`, closing this PR.

PR #2655, opened 2021-04-07
[h2olog] sampling filter with eBPF maps

gfx:
This PR implements a new feature for h2olog: `-S <sampling-rate 0.0-1.0>` by extending h2o's eBPF maps:

Screen Shot 2021-04-07 at 12 01 51

## Restrictions

There're some restrictions in this implementation:

* `-S` flag doesn't work if multiple h2olog processes with `-S` attach to a single h2o process
  * the first one's `-S` is used
* `ebpf_map_lookup(return_map_fd, &tid, &flags)` in `socket.c` could fail because a eBPF's LRU hash map is not guaranteed to have the specified N of entries (and probably the behavior depends on Linux kernel versions)

They are acceptable and thus kept as is for the time being, but could be fixed in the feature. 

## New Configuration

* `usdt-selective-tracing: ON | OFF`
  * This is required to setup eBPF maps used by the `h2olog -S` option

## Others

This PR includes #2669 to write correct tests.


gfx:
It's ready to review!

FYI the changes in `check.mk` is for GitHub Actions, and it is a bit complicated:

* `BPF_MAP_CREATE` fails with `EPERM` if `RLIMIT_MEMLOCK` is too small
* Docker sets `RLIMIT_MEMLOCK` to 65536 by default for containers and it's too small for `BPF_MAP_CREATE`
*  Thus `docker --ulimit $soft:$hard` is needed to configure the limit

There's another solution that h2o calls `setrlimit` before `BPF_MAP_CREATE`. In fact, iovisor/bcc's API does so: https://github.com/iovisor/bcc/blob/master/src/cc/libbpf.c#L282-L294 but I prefer to use docker's option because it's not an h2o issue but the environment issue.

lib/common/socket.c
```diff
@@ -1573,15 +1573,81 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+    if (getuid() != 0) {
+        h2o_error_printf("skipping to set up eBPF maps because bpf(2) requires root privileges\n");
+        return 0;
+    }
 
-static void open_tracing_map(h2o_loop_t *loop)
+    // It creates a pinned BPF object file,
+    // and h2o cannot unlink the file because
+    // h2o drops root privileges after this function
+    // unless its configuration is set to keep the privileges.
```
kazuho, line 1626: Generally speaking we use block comments (i.e., `/* ... */`) to describe what is done by a block, and line-fold at 132 columns, as we do for the source code.

lib/common/socket.c
```diff
@@ -1573,15 +1573,81 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+        return 0;
+    }
+
+    if (ebpf_obj_pin(fd, H2O_EBPF_RETURN_MAP_PATH) != 0) {
+        h2o_perror("BPF_OBJ_PIN failed");
+        return 0;
+    }
+    close(fd); // each worker thread has its own fd
+    return 1;
+}
```
kazuho, line 1640: What does the return code of this function indicate? How should the caller handle the result?

I ask the question because I'm confused; if the return code is a boolean indicating success, then I believe that we should be returning 1 when `ebpf_obj_pin` return an error due to a map already been registered (see my comment above regarding the rationale).

This comment also applies to `h2o_socket_ebpf_setup` function defined in the `#else` part.

I do not have a strong preference regarding which, but think that we might want to do either of the following:
* change the return type to `void`, and exit when an error occurs, or
* change the return type to `int`, clarify what the return code means in the doc-comment of the function declaration (i.e. in the header file), and handle that error code at the call-site.

lib/common/socket.c
```diff
@@ -1573,15 +1573,81 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+    // unless its configuration is set to keep the privileges.
+    int fd =
+        ebpf_map_create(BPF_MAP_TYPE_LRU_HASH, sizeof(pid_t), sizeof(uint64_t), H2O_EBPF_RETURN_MAP_SIZE, H2O_EBPF_RETURN_MAP_NAME);
+    if (fd < 0) {
+        h2o_perror("BPF_MAP_CREATE failed");
+        return 0;
+    }
+
+    if (ebpf_obj_pin(fd, H2O_EBPF_RETURN_MAP_PATH) != 0) {
+        h2o_perror("BPF_OBJ_PIN failed");
```
kazuho, line 1635: We are creating an eBPF map and then calling `ebpf_obj_pin`. What does happen when there is an eBPF map is already registered with the same name? Do we receive an error in that case?

If so, I believe that we should not be reporting an error to the user, as IIUC `get_return_map_fd` function is written to use an existing map.

If not, I believe that we should change the logic to reuse the eBPF map already exists, as the current scheme is designed to allow multiple instances of H2O to use the same return map, and because `get_return_map_fd` is *not* designed to reopen the eBPF map unless it gets removed.

lib/common/socket.c
```diff
@@ -1573,15 +1573,81 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+        return 0;
```
kazuho, line 1636: I believe `close(fd)` is missing here?

lib/common/socket.c
```diff
@@ -1668,39 +1721,100 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
+
+    // map exists, try bpf_obj_get
+    union bpf_attr attr = {
+        .pathname = (uint64_t)H2O_EBPF_RETURN_MAP_PATH,
+    };
+    fd = syscall(__NR_bpf, BPF_OBJ_GET, &attr, sizeof(attr));
+    if (fd < 0)
+        h2o_perror("BPF_OBJ_GET failed");
+    return fd;
+}
```
kazuho, line 1758: It seems to me that `get_tracing_map_fd` and `get_return_map_fd` have identical code with the only difference being the thread-local variables being used and the path of the eBPF map.

I appreciate using consistent approach for opening eBPF maps but wonder if we could avoid copy-pasting. Could we create a common function that does the necessary things for both the eBPF maps?

lib/common/socket.c
```diff
@@ -1668,39 +1721,100 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
+    if ((tracing_map_fd >= 0 || H2O_SOCKET_LOOKUP_FLAGS_ENABLED()) && init_key(&key, cbdata)) {
+        if (tracing_map_fd >= 0)
+            ebpf_map_lookup(tracing_map_fd, &key, &flags);
+
+        int return_map_fd;
+        if (H2O_SOCKET_LOOKUP_FLAGS_ENABLED() && (return_map_fd = get_return_map_fd(loop)) >= 0) {
+            pid_t tid = gettid();
+
+            // make sure a possible old flags is not set,
+            // otherwise the subsequent logic will be unreliable.
```
kazuho, line 1775:
```suggestion
            /* Make sure a possible old flags is not set, otherwise the subsequent logic will be unreliable. */
```
Please see my other comment regarding why.

lib/common/socket.c
```diff
@@ -1668,39 +1721,100 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
+            if (ebpf_map_delete(return_map_fd, &tid) == 0 || errno == ENOENT) {
+                H2O_SOCKET_LOOKUP_FLAGS(tid, flags, &key);
+
+                if (ebpf_map_lookup(return_map_fd, &tid, &flags) != 0) {
+                    if (errno == ENOENT)
```
kazuho, line 1780: Our coding convention is to always use blocks for if-else. Please see https://github.com/h2o/h2o/wiki/Coding-Style.

lib/common/socket.c
```diff
@@ -1668,39 +1721,100 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
+                        // ENOENT could be issued in some reasons even if BPF tries to insert the entry, for example:
+                        //   * the entry in LRU hash was evicted
+                        //   * the insert operation in BPF program failed with ENOMEM
```
kazuho, line 1783:
```suggestion
                        /* ENOENT could be issued in some reasons even if BPF tries to insert the entry, for example:
                        /*  * the entry in LRU hash was evicted
                        /*  * the insert operation in BPF program failed with ENOMEM */
```
This comment is good in sense that it uses bullet points, in which case we should be line-folding before the column limit. But we should be using `/* ... */` rather than `//`.

src/h2olog/h2olog.cc
```diff
@@ -346,6 +368,10 @@ int main(int argc, char **argv)
         probes.push_back(ebpf::USDT(h2o_pid, usdt.provider, usdt.name, usdt.probe_func));
     }
 
+    if (sampling_rate >= 0) {
+        cflags.push_back("-DH2OLOG_SAMPLING_RATE=" + std::to_string(sampling_rate));
+    }
+
```
kazuho, line 374: The tricks around `build_define_cflag` and `CFLAG_D` are fine if not too complex, though I wonder why it is define as such that it cannot be used for adding flags here.

My point is that we might want to convert the nifty tools so that they'd append entries to `std::vector<std::string>`. Then, we could use that for all the `-Dxxx=yyy` flags that we build.

kazuho (commented):
Thank you for working on this. As we discussed elsewhere I think the PR is ready, but rereading this, I think I have some minor comments. PTAL.

src/h2olog/misc/gen_raw_tracer.py
```diff
@@ -423,13 +434,29 @@ def build_tracer(context, metadata):
...
+    c += r"""
+#ifdef H2OLOG_SAMPLING_RATE
+  uint64_t flags = event.socket_lookup_flags.original_flags;
+  int skip_tracing = bpf_get_prandom_u32() > (H2OLOG_SAMPLING_RATE * U32_MAX);
+  if (skip_tracing) {
+    flags |= H2O_EBPF_FLAGS_SKIP_TRACING_BIT;
+  }
+  int64_t ret = h2o_return.insert(&event.socket_lookup_flags.tid, &flags);
+  if (ret != 0)
+    bpf_trace_printk("failed to insert 0x%%lx in %s with errno=%%ld\n", flags, -ret);
```
syohex, line 447: https://www.kernel.org/doc/Documentation/printk-formats.txt

This document says that format specifiers for 64bit data type are `%lld` or `%llu` or `%llx`.

src/h2olog/misc/gen_raw_tracer.py
```diff
@@ -423,13 +434,29 @@ def build_tracer(context, metadata):
 #endif
 """
 
-  c += r"""
+  if fully_specified_probe_name == "h2o:socket_lookup_flags":
+    c += r"""
+#ifdef H2OLOG_SAMPLING_RATE
+  uint64_t flags = event.socket_lookup_flags.original_flags;
+  int skip_tracing = bpf_get_prandom_u32() > (H2OLOG_SAMPLING_RATE * U32_MAX);
```
syohex, line 441: IMHO: eBPF does not support floating point number yet. I suppose that using floating point number in BPF code is not good(in general, floating point number is not used in kernel space). Well, this code does not have problem because a compiler can calculate its value in compile time and generate it as integer value. I suppose percentage(0-100) is better than rate(0.0 - 1.0) as below.

```c
int skip_tracing = bpf_get_prandom_u32() > (U32_MAX / 100 * PERCENTAGE)
```

gfx: Oh, good catch! 

bpf(2) manpage says:

> eBPF programs can be written in a restricted C that is compiled (using the clang compiler) into eBPF bytecode. Various features are omitted from this restricted C, such as loops,  global variables,  variadic functions,  floating-point numbers, and passing structures as function arguments.  Some examples can be found in the samples/bpf/*_kern.c files in the kernel source tree.

I'm not sure the percentage precision is good enough, so I'll calculate `H2OLOG_SAMPLING_RATE * U32_MAX)` in the userspace.

src/h2olog/h2olog.cc
gfx: Actually, we already ready to use `build_define_cflag` here, though I cannot remember why I didn't use it.

Fixed, and I've also renamed `build_define_cflag` to `build_cflag_d` to suggest `CFLAG_D` is a wrapper for `build_cflag_d`.

lib/common/socket.c
gfx: That's a good point. 

I was not sure how we can handle the setup errors, but finally, I'd choose the former: the caller needn't handle errors. For our use case, it is an optional setup, and most of the nodes are not required to setup eBPF stuff. Instead, I've tweaked the warning messages for more helpful to users.

lib/common/socket.c
gfx: I noticed that it just worked even if the pinned map already existed, but in fact `BPF_OBJ_PIN` fails with `EEXIST` in such a case. In that case, however, it seems that the `h2o_socket_ebpf_setup` works as expected, so I've ignored `EEXIST`: 5367b3bace4e52a91682591b24506be4c3e3ed4c

This may be because `BPF_MAP_CREATE` finds an existing map that has the same `attr` first, and only creates a map if it's not found.

lib/common/socket.c
kazuho: > For our use case, it is an optional setup, and most of the nodes are not required to setup eBPF stuff.

Negative side of this approach is that we'd miss when things stop working (due to a bug or a deployment issue). Recent example is #2612.

I might argue that h2o should error-exit when it unexpectedly fails to create the return map. If it is hard to tell when it should succeed (or fail), we can have an option to turn on / off the feature. WDYT?

lib/common/socket.c
```diff
@@ -1668,39 +1733,72 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
+                H2O__PRIVATE_SOCKET_LOOKUP_FLAGS(tid, flags, &key);
+
+                if (ebpf_map_lookup(return_map_fd, &tid, &flags) != 0) {
+                    if (errno == ENOENT) {
+                        /* ENOENT could be issued in some reasons even if BPF tries to insert the entry, for example:
+                         *  * the entry in LRU hash was evicted
+                         *  * the insert operation in BPF program failed with ENOMEM */
+                        h2o_error_printf(
+                            "BPF_MAP_LOOKUP failed. "
+                            "BPF handler for h2o:socket_lookup_flags_private might not have set the flags via h2o_return map\n");
```
kazuho, line 1769: As correctly stated by the comment right above, we expect this branch to be taken at some probability. Based on that, I tend to believe that we should be capping the frequency of this message being emitted. Otherwise, we'd have the chance of this message kicking out all other error messages that would be of more interest.

For a predent, please refer to `track_sendmsg` of lib/http3/common.c

lib/common/socket.c
gfx: Right! It's definitely better to have an option to control this feature. 

How about something like this?

* `ebpf-h2o-return: default | on | off`:
  * Sets up the `h2o_return` eBPF map
  * `default` is "best effort", trying to setup it but ignoring errors. `on` forces to succeed, `off` skips to setup it.

lib/common/socket.c
gfx: Nice catch. 84bec57ea

I'm not sure we'd better implement a shared utility for this kind of logs because we could revert 84bec57ea in the future, the time when we know this ENOENT is unlikely to occur and thus we'd like to know raw logs instead of summarized logs.

lib/common/socket.c
gfx: Talked with @kazuho elsewhere, and we've agreed with `usdt-selective-tracing: ON | OFF`.

gfx:
I think it's ready to re-review!

gfx:
Fixed all the issues so far:

- 489bfd3: Fixed a permission issue by opening a single BPF fd before dropping root privileges, instead of opening an fd per thread (it seems to be safe,  checked with a tiny test program https://github.com/gfx/example-ipc-using-pinned-bpf-maps/commit/76c829eff5b1225405cdfd5730b1e9afca765197).
- https://github.com/gfx/example-ipc-using-pinned-bpf-maps/commit/76c829eff5b1225405cdfd5730b1e9afca765197 is included to test the above commit

lib/common/socket.c
kazuho: Thank you for the changes. Reviewing the added code, I actually noticed that there's an issue in the existing code that's being copy-pasted.

That indicates that it's difficult to implement correctly, therefore I think we should not only fix the bug but also make it reusable so that we do not fall into the same pitfall. Hence filed #2675.

I think we'd prefer merging #2675 then using that as the bases for implementing rate-limiting this warning.

lib/common/socket.c
```diff
@@ -1668,39 +1740,109 @@ int h2o_socket_ebpf_init_key(h2o_ebpf_map_key_t *key, void *_sock)
...
-        return (h2o_ebpf_map_value_t){0};
-    return lookup_map(&key);
+    if ((tracing_map_fd >= 0 || H2O__PRIVATE_SOCKET_LOOKUP_FLAGS_ENABLED()) && init_key(&key, cbdata)) {
+        if (tracing_map_fd >= 0)
+            ebpf_map_lookup(tracing_map_fd, &key, &flags);
+
+        if (H2O__PRIVATE_SOCKET_LOOKUP_FLAGS_ENABLED() && return_map_fd >= 0) {
+            pid_t tid = gettid();
+
+            /* Make sure a possible old flags is not set, otherwise the subsequent logic will be unreliable. */
```
kazuho, line 1793:
```suggestion
            /* Make sure old flags do not exist, otherwise the subsequent logic will be unreliable. */
```

lib/core/configurator.c
```diff
@@ -958,6 +958,15 @@ static int on_config_stash(h2o_configurator_command_t *cmd, h2o_configurator_con
     return 0;
 }
 
+static int on_config_usdt_selective_tracing(h2o_configurator_command_t *cmd, h2o_configurator_context_t *ctx, yoml_t *node)
+{
+    ssize_t on;
+    if ((on = h2o_configurator_get_one_of(cmd, node, "OFF,ON")) == -1)
+        return -1;
+    ctx->globalconf->usdt_selective_tracing = (uint8_t)on;
```
kazuho, line 966:
```suggestion
    ctx->globalconf->usdt_selective_tracing = (int)on;
```

Assuming that the intent is to use `int` as being defined in the header file.

src/h2olog/h2olog.cc
```diff
@@ -226,6 +225,18 @@ static void lost_cb(void *context, uint64_t lost)
     tracer->handle_lost(lost);
 }
 
+template <typename T> static std::string build_cflag_d(const char *name, const T &value)
+{
+    return std::string("-D") + std::string(name) + "=" + std::to_string(value);
```
kazuho, line 230:
```suggestion
    return build_cflag_d(name, std::to_string(value).c_str());
```

As an alternative, we can omit `.c_str()` if we change the second argtype of the other function definition from `const char *` to `const std::string &`.

src/h2olog/h2olog.cc
```diff
@@ -226,6 +225,18 @@ static void lost_cb(void *context, uint64_t lost)
     tracer->handle_lost(lost);
 }
 
+template <typename T> static std::string build_cflag_d(const char *name, const T &value)
```
kazuho, line 228: Maybe call these `define_cc_macro` or something? `_d` sounds like a bit confusing confusing.

kazuho (commented):
Thank you for the changes. I have left comments, PTAL.

lib/common/socket.c
```diff
@@ -1573,56 +1573,128 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+                             "maybe because RLIMIT_MEMLOCK is too small.\n");
+        } else {
+            h2o_perror("BPF_MAP_CREATE failed");
+        }
+        goto Exit;
+    }
+
+    if (ebpf_obj_pin(fd, H2O_EBPF_RETURN_MAP_PATH) != 0) {
+        if (errno == EEXIST) {
+            success = 1;
```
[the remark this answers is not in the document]
gfx: Ugh...in fact, I misunderstood it, the case where `H2O_EBPF_RETURN_MAP_PATH` already exists. Even the test was wrong. 

If the map already exists, we have to open the file with `BPF_OBJ_GET`, making sure the map's type and name are correct. Fixed dc20115ed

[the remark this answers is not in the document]
gfx: FWIW In case the map file already exists, h2o checks its critical attributes, namely type, key size, value size (626a3f448). This check is not to detect attacks but for the changes in the future.  When we changed these attributes in the future, we have to remove the pinned map file on deployment.

src/h2olog/h2olog.cc
gfx: Refactored these stuff: d27775e6bf48707d47e02700f3051cdcf5ec3c14

Before the above commit, `build_cflag_d(name, T)` and `build_cflag_d(name, std::string)` behaved differently: the former makes `-D$name=$expr` and the latter makes `-D$name="$expr"`. I think this was a kind of overloading abuse. So I've splitted it to `build_cc_macro_expr()` and `build_cc_macro_str()`, and now there's no implementation redundancy.

lib/common/socket.c
```diff
@@ -1570,59 +1570,196 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+    };
+
+    int success = 0;
+    int fd = -1;
+    if (getuid() != 0) {
+        h2o_error_printf("failed to set up eBPF maps because bpf(2) requires root privileges\n");
+        goto Exit;
+    }
+
+    if (!file_exist(H2O_EBPF_RETURN_MAP_PATH)) {
```
kazuho, line 1665: Why don't we call `ebpf_obj_pin` directly?

If the call succeeds, then we'd be all set.

If it returns `ENOENT`, then we can go on to creating a new one.

lib/common/socket.c
```diff
@@ -1570,59 +1570,196 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+        if (m.type != map_attr.type) {
+            h2o_error_printf("Unexpected map type: expected %d but got %d\n", map_attr.type, m.type);
+            goto Exit;
+        }
+        if (m.key_size != map_attr.key_size) {
+            h2o_error_printf("Unexpected map key size: expected %" PRIu32 " but got %" PRIu32 "\n", map_attr.key_size, m.key_size);
+            goto Exit;
+        }
+        if (m.key_size != map_attr.key_size) {
+            h2o_error_printf("Unexpected map value size: expected %" PRIu32 " but got %" PRIu32 "\n", map_attr.value_size,
```
kazuho, line 1711: Regarding these error messages, I think we might be clear that the "map" we are talking about is the specific ebpf map (i.e. H2O_EBPF_RETURN_MAP_PATH) that we are dealing with. Otherwise, "map" is such a generic term that users might have difficulty in finding out the cause.

lib/common/socket.c
```diff
@@ -1570,59 +1570,196 @@ void h2o_sliding_counter_stop(h2o_sliding_counter_t *counter, uint64_t now)
...
+        .key = (uint64_t)key,
+    };
+    return syscall(__NR_bpf, BPF_MAP_DELETE_ELEM, &attr, sizeof(attr));
+}
+
+static int file_exist(const char *path)
+{
+    struct stat s;
+    return stat(path, &s) == 0;
+}
```
kazuho, line 1642: I think we might prefer not creating this function, because the only caller would be `get_map_fd` (assuming that we adopt my other review comment), and because the code is not testing if a "file_exist" (when we consider the directory permissions).

kazuho, line 1642: I think we might prefer not creating this function, because the only caller would be `get_map_fd` (assuming that we adopt my other review comment), and because the code is not testing if a "file_exist" (when we consider the directory permissions).

src/h2olog/h2olog.cc
```diff
@@ -226,6 +225,30 @@ static void lost_cb(void *context, uint64_t lost)
     tracer->handle_lost(lost);
 }
 
+/**
+ * Builds a `-D$name=$expr` style cc macro.
+ */
+std::string build_cc_macro_expr(const char *name, const std::string &expr)
```
kazuho, line 231:
```suggestion
static std::string build_cc_macro_expr(const char *name, const std::string &expr)
```

Maybe?

kazuho (commented):
Thank you for the changes. I have some minor comments. PTAL.

gfx:
Thank you for the review. Finished all the suggested changes and waiting for CI.

kazuho:
Thank you for all your efforts. Merged to master.

gfx:
Yey! Thank you!

PR #2680, opened 2021-04-30
[h2olog] stop filtering quicly:debug_message out (+ some code cleanups)

gfx:
`quicly:debug_message` (guts of `quicly_debug_printf()`) is not a tracing event but useful for debugging, so there's no need to filter it out. 

Plus, removed some unused variables in h2olog.


kazuho (commented):
I'm not entirely sure if we want to merge this, because what this PR does is "emit debugging messages by default."

The benefit of a debug-message probe is that while testing some code, you can add arbitrary logging code, then selectively turn it on by attaching a probe. OTOH, h2olog is a tool that is used for purposes other than development, and therefore, emission of debug log can be harmful depending on the amount of the message.

FWIW, I would not oppose to adding a command line flag to `h2olog` that, when specified, turns on the debug-message probe. But that's not high in my ToDo list as there are other ways of attaching a debugging probe (e.g., bpftrace).

gfx:
I see. I think I'll use the debug print messages for development, but as you suggested we can do it without h2olog. 

Will make a PR for the refactoring, and (probably) will make another one to explain how to see `quicly:debug_message` with bpftrace.

PR #2690, opened 2021-05-09
fix attribute validations for eBPF maps

gfx:
amending https://github.com/h2o/h2o/pull/2655

Thanks to @hfujita for pointing it out (https://github.com/h2o/h2o/commit/626a3f44826746cfc3c377ca85928d38d3f8dc19#r50571561)


hfujita (approved):
👍 Thank you for the quick fix!

PR #2695, opened 2021-05-17
[fuzzer] Set proxy timeouts

hfujita:
`connect_timeout` and `first_byte_timeout` need to be explicitly
set to non-zero values.
This patch aligns these values with `io_timeout`, as done in
https://github.com/h2o/h2o/blob/35401aeb71bd9b8f97e543c524aae50af24e0978/lib/handler/configurator/proxy.c#L46


deweerdt (approved):
Thank you!

hfujita:
Due to increase in proxy io timeout, http1 fuzzer run in CI (executes corpora under `fuzz/http1-corpus`) now takes 75 seconds, which used to take 45 seconds. Hopefully this is still acceptable.

hfujita:
@kazuho I think this is ready to merge. Let me know what you think.

PR #2718, opened 2021-05-28
Do not start reading on tunnel since h2o_send will call tcp_on_proceed

deweerdt:
The `h2o_socket_read_start` is unnecessary because we immediately call `h2o_send`,
which will call `tcp_on_proceed`, which will set the socket in a read
state.

It's a bug to have start reading because we might have
`tcp_on_read` called on its own right before `tcp_on_proceed`
is called, and we will assert in `http2::finalostream_send` in
`h2o_http2_conn_register_for_proceed_callback` because the stream is
already linked


kazuho (approved):

PR #2719, opened 2021-05-28
h2olog: handle all the CID fields in addition to master_id

gfx:
To collect logs by ID, each event that is associated with `quicly_conn_t` must have enough information about the connection id. 

This is an idea: expanding `quicly_cid_plaintext_t` to a JSON array with `[master_id,  thread_id, path_id, node_id]`. A log collector may use only `master_id` and `thread_id` to identify the QUIC connection.

I've also added a test to make sure `qlog-adapter.py` can process h2olog output.


gfx:
Closing in favor of #2731.

PR #2721, opened 2021-05-28
[fuzzer] Use timeout specified by envvar

hfujita:
H2O_FUZZER_CLIENT_TIMEOUT is read from environment but never used.
This patch actually uses it so the behavior matches what README says.


kazuho (approved):

PR #2737, opened 2021-06-02
h2olog: remove `time` field added by h2olog since it's confusing and useless

gfx:
IIRC `time` field should've been removed when the `seq` field was introduced for sorting. It's just confusing and even misleading. If we need `time` in events from `h2o-probes.d`, we should add `time` to each USDT.


kazuho (commented):
I'm not sure if I follow.

Don't we need `time` to recover when each event happened? I'm not sure if we'd want to log things without timestamps.

gfx:
* Each event should be timestamped
* It may be better for each `h2o:*` event  to have h2o internal time when `qlog-adapter.py` processes h3 events as well

PR #2744, opened 2021-06-07
[h2olog] add timestamp to each hX_accept to identify a particular HTTP connection

gfx:
To identify an HTTP (not QUIC) connection with `hX_accept`, I need a timestamp for each `hX_accept` event. In practice, `$hostname-$h2o_pid-$h2o_conn_id-$connected_at` is the identifier.

BTW I'm adding `h2_accept` and `h2_close` with the same fields with H1's just for consistency.

## example row

```
{"type":"h3s-accept","tid":18833,"seq":13,"conn-id":2,"conn":107133664502144,"quic-master-id":0,"connected-at":1623032571650185,"time":1623032571656}
```

### DIFF
--- a/h2o-probes.d
+++ b/h2o-probes.d
@@ -62,12 +62,20 @@ provider h2o {
     /**
      * HTTP/1 server-level event, indicating that a connection has been accepted.
      */
-    probe h1_accept(uint64_t conn_id, struct st_h2o_socket_t *sock, struct st_h2o_conn_t *conn);
+    probe h1_accept(uint64_t conn_id, struct st_h2o_socket_t *sock, struct st_h2o_conn_t *conn, struct timeval connected_at);
     /**
      * HTTP/1 server-level event, indicating that a connection has been closed.
      */
     probe h1_close(uint64_t conn_id);
 
+    /**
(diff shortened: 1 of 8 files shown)


h2o-probes.d
```diff
@@ -62,12 +62,20 @@ provider h2o {
     /**
      * HTTP/1 server-level event, indicating that a connection has been accepted.
      */
-    probe h1_accept(uint64_t conn_id, struct st_h2o_socket_t *sock, struct st_h2o_conn_t *conn);
+    probe h1_accept(uint64_t conn_id, struct st_h2o_socket_t *sock, struct st_h2o_conn_t *conn, struct timeval *connected_at);
```
gfx, line 65: FWIW I use a pointer for `struct timeval` just because BCC is not able to handle a 128-bit value object, which tires to use `uint128_t` internally, but there's no `uint128_t` in LLVM-BPF.

gfx:
Talked with @kazuho in elsewhere, and we've agreed that we can identify an HTTP connection with hostname, pid, and [time_of_process_creation](https://unix.stackexchange.com/questions/62154/when-was-a-process-started). No need to add `connected_at` for this purpose. Closing.

PR #2745, opened 2021-06-08
[h2olog] add connection UUID in the  `h2o_conn_t* conn` param 

gfx:
This PR adds a field `char uuid[37]` to `h2o_conn_t`  (available via `h2o_conn_get_uuid()`) to identify each HTTP connection. This is particularly useful for h2olog, which now identify each HTTP connection in a consistent way. 

<del>This feature is disabled by default since `h2o_generate_uuid()` consumes more resources and memory for each HTTP connection.</del> <ins>The connection UUID is only initialized when `h2o_conn_get_uuid()` is called so no CPU resources are consumed if it's not used</ins>.

example output:

```json
{"type":"h3s-accept","tid":147324,"seq":13,"conn-id":2,"conn-uuid":"0f1e527a-cc99-4400-0053-8244570e76d0","quic-master-id":0,"time":1623118531730}
```

## Refs

* https://datatracker.ietf.org/doc/html/rfc4122

### DIFF
--- a/lib/common/rand.c
+++ b/lib/common/rand.c
@@ -26,8 +26,18 @@
 #include "h2o/string_.h"
 #include "picotls/openssl.h"
 
-static void format_uuid_rfc4122(char *dst, const uint8_t *uuid)
+static void format_uuid_rfc4122(char *dst, uint8_t *uuid, uint8_t version)
 {
+    // Variant:
+    // > Set the two most significant bits (bits 6 and 7) of the
+    // > clock_seq_hi_and_reserved to zero and one, respectively.
+    uuid[8] = (uuid[8] & 0x3f) | 0x80;
+    // Version:
+    // > Set the four most significant bits (bits 12 through 15) of the
+    // > time_hi_and_version field to the 4-bit version number from
+    // > Section 4.1.3.
+    uuid[6] = (uuid[6] & 0x0f) | (version << 4);
+
     // String Representation:
     // > UUID  = time-low "-" time-mid "-"
     // >         time-high-and-version "-"
@@ -65,16 +75,5 @@ void h2o_generate_uuidv4(char *buf)
 
     uint8_t uuid[16];
     ptls_openssl_random_bytes((void *)&uuid, sizeof(uuid));
-
-    // Variant:
-    // > Set the two most significant bits (bits 6 and 7) of the
-    // > clock_seq_hi_and_reserved to zero and one, respectively.
-    uuid[8] = (uuid[8] & 0x3f) | 0x80;
-    // Version:
-    // > Set the four most significant bits (bits 12 through 15) of the
-    // > time_hi_and_version field to the 4-bit version number from
-    // > Section 4.1.3.
-    uint8_t version = 4;
-    uuid[6] = (uuid[6] & 0x0f) | (version << 4);
-    format_uuid_rfc4122(buf, uuid);
+    format_uuid_rfc4122(buf, uuid, 4);
 }
--- a/t/00unit/lib/common/rand.c
+++ b/t/00unit/lib/common/rand.c
@@ -24,31 +24,20 @@
 
 static void test_format_uuid_rfc4122(void)
 {
-    uint8_t z[16] = {0};
+    uint8_t uuid[16] = {0};
     char dst[H2O_UUID_STR_RFC4122_LEN + 1];
-    format_uuid_rfc4122(dst, z);
 
+    format_uuid_rfc4122(dst, uuid, 4);
     ok(strlen(dst) == H2O_UUID_STR_RFC4122_LEN);
-    ok(strcmp(dst, "00000000-0000-0000-0000-000000000000") == 0);
-}
+    ok(strcmp(dst, "00000000-0000-4000-8000-000000000000") == 0);
 
-static void test_h2o_generate_uuidv4(void)
-{
-    for (int i = 0; i < 10; i++) {
-        char dst[H2O_UUID_STR_RFC4122_LEN + 1];
-        h2o_generate_uuidv4(dst);
-
-        ok(strlen(dst) == H2O_UUID_STR_RFC4122_LEN);
-        // https://stackoverflow.com/a/19989922/805246
-        // > Version 4 UUIDs have the form xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
-        // > where x is any hexadecimal digit and y is one of 8, 9, A, or B.
-        ok(dst[14] == '4');
-        ok(dst[19] == '8' || dst[19] == '9' || dst[19] == 'a' || dst[19] == 'b');
-    }
+    memset(uuid, 0xff, sizeof(uuid));
+    format_uuid_rfc4122(dst, uuid, 4);
+    ok(strlen(dst) == H2O_UUID_STR_RFC4122_LEN);
+    ok(strcmp(dst, "ffffffff-ffff-4fff-bfff-ffffffffffff") == 0);
 }
 
 void test_lib__common__rand_c(void)
 {
     subtest("format_uuid_rfc4122", test_format_uuid_rfc4122);
-    subtest("h2o_generate_uuidv4", test_h2o_generate_uuidv4);
 }


include/h2o.h
```diff
@@ -961,6 +971,13 @@ struct st_h2o_conn_t {
      * callbacks
      */
     const h2o_conn_callbacks_t *callbacks;
+
+#if H2O_USE_CONN_UUID
+    /**
+     * connection UUID (UUIDv4 string representation)
+     */
+    char uuid[37];
+#endif
```
kazuho, line 980: I did say elsewhere that this should be a compile-time option. Let me change my position.

I think that UUID should always be supported, but lazily-initialized, so that it would consume CPU cycles only when used. 37 bytes (or possibly 38, if we add a flag to indicate if the UUID has been generated) is nothing compared to TCP send / receive window size.

For regular usage, we can create a function called `h2o_conn_get_uuid` that returns UUID, generating on the fly it is has not yet been generated. In case of probe points, we can explicitly call that function, with a comment stating that it has to be explicitly created as the attached probe reads the memory directly.

WDYT?

lib/core/util.c
```diff
@@ -945,3 +946,60 @@ void h2o_cleanup_thread(void)
...
+        uint16_t time_hi_and_version;
+        uint8_t clock_seq_hi_and_reserved;
+        uint8_t clock_seq_low;
+        uint8_t node[6];
+    } uuid;
+    _Static_assert(sizeof(uuid) == 16, "size of UUID structure must be 16");
+
+    if (RAND_bytes((void*)&uuid, sizeof(uuid)) != 1) {
+        h2o_fatal("RAND_bytes failed");
+    }
```
kazuho, line 967: This is more of an editorial choice, but please use `ptls_openssl_random_bytes` instead, as we do elsewhere.

lib/core/util.c
```diff
@@ -945,3 +946,60 @@ void h2o_cleanup_thread(void)
...
+    // > Set the two most significant bits (bits 6 and 7) of the
+    // > clock_seq_hi_and_reserved to zero and one, respectively.
+    uuid.clock_seq_hi_and_reserved = (uuid.clock_seq_hi_and_reserved & 0x3f) & 0x80;
+
+    // Version:
+    // > Set the four most significant bits (bits 12 through 15) of the
+    // > time_hi_and_version field to the 4-bit version number from
+    // > Section 4.1.3.
+    const int version = 4;
+    uuid.time_hi_and_version = (uuid.time_hi_and_version & 0x0f) | (version << 4);
```
kazuho, line 979: Is this correct?

I believe that the value is expected to preserve 12 bits of random with remaining 4 bits used to represent the version number. But my reading of this statement is that only 4 bits of random is being preserved.

I appreciate your efforts to follow what the specification says, but I wonder if it would actually be better to simply do everything flattened all the time; i.e. generate random bytes in a byte array (`uint8_t [16]`), flip the necessary bits, then serialize it.

src/h2olog/misc/gen_raw_tracer.py
```diff
@@ -76,6 +76,10 @@
         ("super.local.cid_set.plaintext.master_id", "master_id"),
     ]],
 
+    ["st_h2o_conn_t", [
+        ("uuid", None),
+    ]],
```
kazuho, line 81: I think we'd prefer retaining the capability of logging conn_id (of 64-bits) as well.

kazuho (commented):
Thank you for the PR. I think we are heading towards the right direction.

include/h2o.h
gfx: It sounds much simpler than compile-time option. Added `h2o_conn_get_uuid()`!

lib/core/util.c
gfx: I thought using struct was much simpler than `uint8_t[16]`, but it has turned out you are right -- I was confused by the byte order.

src/h2olog/misc/gen_raw_tracer.py
gfx: `conn_id` is passed separately, but I've reverted the changes in `gen_raw_tracer.py` anyway.

lib/common/rand.c
```diff
@@ -0,0 +1,55 @@
...
+    UUID_ENC_PART(dst, pos, uuid, 6, 7); /* time_hi_and_version */
+    dst[pos++] = '-';
+    UUID_ENC_PART(dst, pos, uuid, 8, 8); /* clock_seq_hi_and_reserved */
+    UUID_ENC_PART(dst, pos, uuid, 9, 9); /* clock_seq_low */
+    dst[pos++] = '-';
+    UUID_ENC_PART(dst, pos, uuid, 10, 15); /* node */
+#undef UUID_ENC_PART
+
+    /* '\0' is set by h2o_hex_encode() */
+}
```
kazuho, line 34: If you want to have unit test for this function, you can create t/00unit/lib/common/rand.c, much like other files under t/00unit/lib/common.

lib/common/rand.c
```diff
@@ -0,0 +1,55 @@
+#include <openssl/rand.h>
```
kazuho, line 1: Please add the MIT copyright boilerplate, with year set to 2021 and the copyright owners set to the people who have written the code (or who own the right).

kazuho (commented):
Thank you for the changes.

lib/http1.c
```diff
@@ -1337,7 +1337,7 @@ void h2o_http1_accept(h2o_accept_ctx_t *ctx, h2o_socket_t *sock, struct timeval
     sock->data = conn;
     h2o_linklist_insert(&ctx->ctx->http1._conns, &conn->_conns);
 
-    H2O_PROBE_CONN(H1_ACCEPT, &conn->super, conn->sock, &conn->super);
+    H2O_PROBE_CONN(H1_ACCEPT, &conn->super, conn->sock, &conn->super, h2o_conn_get_uuid(&conn->super));
```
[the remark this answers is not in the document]
gfx: That's what `H2O_PROBE*()` does: https://github.com/h2o/h2o/blob/909da44bf32d7da1ad64cea871badabcffda34f5/lib/probes_.h#L48-L54

> BTW, do we need to pass each member of conn when we also pass h2o_conn_t * as a pointer? I'm not requesting a change, though I thought that you created a mechanism for extracting multiple members of a struct. Hence asking.

I think no. For example `uint64_t conn_id` in `h2o:h3s_accept` probe, h2olog doesn't need it. It can be extracted from `h2o_conn_t *conn`. However, it might be convenient for other dtrace tools, although I'm not sure the use-case when h2olog gets production-ready.

lib/common/rand.c
```diff
@@ -0,0 +1,55 @@
...
+static void format_uuid_rfc4122(char *dst, const uint8_t *uuid)
+{
+    // String Representation:
+    // > UUID  = time-low "-" time-mid "-"
+    // >         time-high-and-version "-"
+    // >         clock-seq-and-reserved
+    // >         clock-seq-low "-" node
+    // See also "4.1.2. Layout and Byte Order" for the layout
+    size_t pos = 0;
+#define UUID_ENC_PART(b, p, u, start, last) do { \
```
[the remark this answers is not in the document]
gfx: > Considering the fact that this is a closure, maybe use dst, pos, uuid of the enclosing function rather than doing pass-by-name as b, p, u?

✅ done

---

For the other parts, I agree with what you pointed. In fact, `start` and `last` is not matched to the C/C++ convention, and `start - last + 1` to calculate the size seems a bit strange to me as well, but I'd like to keep them for that macro because it is meant to exactly matched to what the spec says in the 4.1.2. "Layout and Byte Order".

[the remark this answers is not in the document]
kazuho: I see. If that's the concern, I'd suggest using "first" and "last" as the terms. Because "last X" implies that X exists. That's the different from using "end."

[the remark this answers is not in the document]
gfx: That's great! Changed to use the terms "first" and "last" 517af66efc43687c4fe14f28cc3b0f95f76ce737

include/h2o.h
```diff
@@ -961,6 +961,14 @@ struct st_h2o_conn_t {
      * callbacks
      */
     const h2o_conn_callbacks_t *callbacks;
+
```
kazuho, line 964:
```suggestion
```
I do not think there's a reason to add an empty line here, when there's none between the members above?

include/h2o.h
```diff
@@ -2203,11 +2215,21 @@ inline h2o_conn_t *h2o_create_connection(size_t sz, h2o_context_t *ctx, h2o_host
 #else
     conn->id = __sync_add_and_fetch(&h2o_connection_id, 1);
 #endif
+    conn->_uuid.is_initialized = 0;
     conn->callbacks = callbacks;
```
kazuho, line 2219:
```suggestion
    conn->callbacks = callbacks;
    conn->_uuid.is_initialized = 0;
```

Let's do things in the order they are defined.

lib/common/rand.c
```diff
@@ -0,0 +1,78 @@
...
+
+static void format_uuid_rfc4122(char *dst, const uint8_t *uuid)
+{
+    // String Representation:
+    // > UUID  = time-low "-" time-mid "-"
+    // >         time-high-and-version "-"
+    // >         clock-seq-and-reserved
+    // >         clock-seq-low "-" node
+    // See also "4.1.2. Layout and Byte Order" for the layout
+    size_t pos = 0;
```
kazuho, line 37:
```suggestion
    size_t pos = 0;

```
If we are to add empty lines before `#define` and `#undef`, then I'd say that we'd better have empty lines before them as well.

lib/common/rand.c
```diff
@@ -0,0 +1,78 @@
...
+    UUID_ENC_PART(0, 3); /* time_low */
+    dst[pos++] = '-';
+    UUID_ENC_PART(4, 5); /* time_mid */
+    dst[pos++] = '-';
+    UUID_ENC_PART(6, 7); /* time_hi_and_version */
+    dst[pos++] = '-';
+    UUID_ENC_PART(8, 8); /* clock_seq_hi_and_reserved */
+    UUID_ENC_PART(9, 9); /* clock_seq_low */
+    dst[pos++] = '-';
+    UUID_ENC_PART(10, 15); /* node */
```
kazuho, line 53:
```suggestion
    UUID_ENC_PART(10, 15); /* node */

```

lib/http1.c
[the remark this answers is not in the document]
kazuho: > That's what `H2O_PROBE*()` does: 

Point taken. Yeah it looks like a function call, but the arguments are actually evaluated lazily, because it is a macro.

>  However, it might be convenient for other dtrace tools, although I'm not sure the use-case when h2olog gets production-ready.

Yeah, any dtrace tools can read values given a pointer. We already rely on that for things like `quicly_rtt_t`. That said, I'm happy to leave this issue to another PR (of using multiple arguments for passing members of `h2o_conn_t` while passing `h2o_conn_t` as a pointer), if that's your preference.

lib/common/rand.c
```diff
@@ -0,0 +1,78 @@
...
+    // > Set the two most significant bits (bits 6 and 7) of the
+    // > clock_seq_hi_and_reserved to zero and one, respectively.
+    uuid[8] = (uuid[8] & 0x3f) | 0x80;
+    // Version:
+    // > Set the four most significant bits (bits 12 through 15) of the
+    // > time_hi_and_version field to the 4-bit version number from
+    // > Section 4.1.3.
+    uint8_t version = 4;
+    uuid[6] = (uuid[6] & 0x0f) | (version << 4);
+    format_uuid_rfc4122(buf, uuid);
```
kazuho, line 77:
```suggestion
    uint8_t random_bytes[16];
    ptls_openssl_random_bytes((void *)random_bytes, sizeof(random_bytes));
    format_uuid_rfc4122(buf, random_bytes);
```

I'd say that this function should better be as simple as this, and that the bit-flipping logic should go into `format_uuid_rfc4122`.

Then, you can pass whatever value (e.g., 0xffffff...ffff) to `format_uuid_rfc4122` and see if the function sets the special bits as necessary.

t/00unit/lib/common/rand.c
```diff
@@ -0,0 +1,54 @@
...
+        h2o_generate_uuidv4(dst);
+
+        ok(strlen(dst) == H2O_UUID_STR_RFC4122_LEN);
+        // https://stackoverflow.com/a/19989922/805246
+        // > Version 4 UUIDs have the form xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
+        // > where x is any hexadecimal digit and y is one of 8, 9, A, or B.
+        ok(dst[14] == '4');
+        ok(dst[19] == '8' || dst[19] == '9' || dst[19] == 'a' || dst[19] == 'b');
+    }
+}
```
kazuho, line 48: If you follow my comment above regarding refactoring, we no longer need this test that tries to detect issues in a probabilistic manner.

t/90h2olog.t
```diff
@@ -64,6 +64,9 @@ subtest "h2olog", sub {
 
   my @events = map { decode_json($_) } split /\n/, $trace;
   is scalar(grep { $_->{type} && $_->{tid} && $_->{seq} } @events), scalar(@events), "each event has type, tid and seq";
+
+  my($h3s_accept) = grep { $_->{type} eq "h3s-accept" } @events;
+  ok is_uuidv4($h3s_accept->{"conn-uuid"}), "h3s-accept has a UUIDv4 field `conn-uuid`"
```
kazuho, line 69: Maybe we don't need this, now that we have unit tests for UUID generation. We do not test each value provided through h2olog, do we?

gfx: I'd like to keep the E2E test in addition to unit tests because (1) it's a key to a log collector that uses h2olog, and (2) it's not trivial to handle a string field in h2olog.

lib/http1.c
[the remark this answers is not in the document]
gfx: Agreed. I'm willing to tackle with that redundancy 😄

lib/common/rand.c
gfx: I was hesitating for `format_uuid_rfc4122()` to mutate its input `uuid`, but yeah, that makes `format_uuid_rfc4122` more testable in a robust and simple way.  Fixed.

t/90h2olog.t
kazuho: Works for me.

kazuho (approved):
Thank you for all the changes. LGTM.

PR #2754, opened 2021-06-13
fix bootstrap stall on session ticket file failure

i110:
Currently QUIC configuration causes bootstrap stall if invalid or non-existent session ticket file is provided. This PR fixes it along with a new test case for invalid session ticket file.


i110:
As discussed offline, QUIC must have the session ticket key at bootstrap time. In that sense stalling is the expected behavior.

PR #2771, opened 2021-06-24
Support setting SO_MARK on proxy sockets (TCP and UDP).

jhatala:
New configuration commands:

1. `proxy.connect.socket-mark`: The SO_MARK value (in hexadecimal) to apply to the backend socket for CONNECT and CONNECT-UDP requests.  This requires the Linux capability `CAP_NET_ADMIN` (see the command `capabilities`).

Example:
```
proxy.connect.socket-mark: 0x200
```

2. `capabilities`: a list of Linux capability names that the h2o process will retain after switching to the user specified by the `user` command.

Example:
```
capabilities: ["CAP_NET_ADMIN"]
```

The example configuration file `examples/h2o/h2o-connect-somark.conf` can be used to exercise this functionality.


jhatala:
This will be pullrequestized differently, closing.

PR #2775, opened 2021-06-29
t/50connect-deadlock flakyness, make the test wait longer for read and writes

deweerdt:


deweerdt:
ah! The test falked on this PR, so clearly, something else is going on. Will close

PR #2779, opened 2021-06-30
add h2o_mem_calloc to memory.h

robguima:
Does it make sense to add `h2o_mem_calloc` for convenience, even if there is no use of it yet in the code?

Thanks!

### DIFF
--- a/include/h2o/memory.h
+++ b/include/h2o/memory.h
@@ -196,6 +196,10 @@ static h2o_iovec_t h2o_iovec_init(const void *base, size_t len);
  */
 H2O_RETURNS_NONNULL static void *h2o_mem_alloc(size_t sz);
 /**
+ * wrapper of calloc; allocates and sets to zero given size of memory or dies if impossible
+ */
+H2O_RETURNS_NONNULL static void *h2o_mem_calloc(size_t nmemb, size_t sz);
+/**
  * warpper of realloc; reallocs the given chunk or dies if impossible
  */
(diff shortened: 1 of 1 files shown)


kazuho (commented):
Thank you for the PR. Would you mind elaborating how we might use calloc?

Generally speaking, I think we should allocate memory using `malloc` then initialize all the members, rather than allocating a zero-cleared memory then overwriting the values that have to be something other than zero.

The rationale is that the former is easy to write, more explicit, and faster to execute (consider store forwarding stalls) assuming that the allocation size is small enough to not fallback to `mmap`, i.e.:
```c
p = malloc(sizeof(*p));
p = (struct T){...};
```

robguima:
> Thank you for the PR. Would you mind elaborating how we might use calloc?
> 
> Generally speaking, I think we should allocate memory using `malloc` then initialize all the members, rather than allocating a zero-cleared memory then overwriting the values that have to be something other than zero.
> 
> The rationale is that the former is easy to write, more explicit, and faster to execute (consider store forwarding stalls) assuming that the allocation size is small enough to not fallback to `mmap`, i.e.:
> 
> ```c
> p = malloc(sizeof(*p));
> p = (struct T){...};
> ```

thank you yes, that makes sense. I think the use case where we use calloc the most is when zero-initializing all fields initially and then testing the zero values for late initializations. Especially in our use case where we may have several late initializations taking place when things are ready to load.

I will close the PR and keep it local, np. Thank you for the input.

PR #2847, opened 2021-09-27
http3 fuzzer fixes

i110:
This fixes 2 minor issues of h2o-fuzzer-http3 which I faced on macOS.

- `malloc.h` is Linux specific and we even don't use malloc and others in this program
- Couldn't find `<openssl/openssl.h>` without adding $OPENSSL_INCLUDE_DIR as include directory


kazuho (approved):
Nice changes. Thank you for the PR.

hfujita (approved):
Thank you for the fixes! Somehow macOS support got completely dropped from my head.

PR #2888, opened 2021-11-23
add current directory to @INC in upstream.psgi

i110:
As of Perl 5.26 current directory is not included in `@INC`, so plackup with recent perl fails with `module t::Util not found` error


i110:
`$ENV{PERL5LIB}` is the best

PR #2923, opened 2022-01-17
QUIC packet received/processed counters

sharksforarms:
This PR introduces counters for received quic packets and processed quic packets. This can be useful to monitor for mis-configurations/attacks/etc if the counters diverge greatly.

`quic.packet-received` and `quic.packet-processed` in the json metrics.

I included a test asset which is a initial quic packet with an invalid token. I generated this locally using the h2o spawned from the test config, made a request in firefox and grabbed the quic data from wireshark. I then modified the last byte of the token to invalidate it.


kazuho (commented):
Thank you for the PR.

PR #2930, opened 2022-01-28
update curl to 7.81.0

i110:
@miyagawa found that `t/50fastcgi.t` doesn't require curl 7.57 to pass, so we have no reason to use old curl which was released 4 years ago.

The latest curl requires nghttp2 >= 1.12.0, so I tweaked installation order a bit for ubuntu 16.04.

Tentatively I pushed the updated images with tags `ubuntu1604-curl-7_81_0` and `ubuntu1604-curl-7_81_0` and tested. If the changes look good, I'll push them without suffixes and test again (or, do we have better way of updating images?)


kazuho (approved):
@i110 Thank you for the PR. Changes LGTM. Please update the CI images and merge this PR.

i110:
I rebuilt and pushed images with tags `ubuntu1604` and `ubuntu200`.

PR #2931, opened 2022-02-01
Draft: Use Util::empty_ports to avoid listening port which already in used.

u5surf:
Occasionally, CI failed following reason.

https://github.com/h2o/h2o/runs/5013040895?check_suite_focus=true
```
# t/50proxy-forward-close-connection.t --------------------------------
  0.201992 # Subtest: basic
  0.202423     # Subtest: enabled
  0.216902 spawning /home/ci/h2o... [INFO] raised RLIMIT_NOFILE to 1024
  0.222665 h2o server (pid:22851) is ready to serve requests with 2 threads
  0.251843 fetch-ocsp-response (using OpenSSL 1.0.2g  1 Mar 2016)
  0.255133 failed to extract ocsp URI from /tmp/5n3OL9Pho0/cert.crt
  0.260409 [OCSP Stapling] disabled for certificate file:examples/h2o/server.crt
  0.306010 done
  0.306022 killing /home/ci/h2o... received SIGTERM, gracefully shutting down
  0.406118 killed (got 0)
  0.406505         1..0
  0.406667         # No tests run!
  0.407083     not ok 1 - No tests run for subtest "enabled"
  0.407254     #   Failed test 'No tests run for subtest "enabled"'
  0.407256     #   at t/50proxy-forward-close-connection.t line 44.
  0.407573     1..1
  0.407660     # Looks like you failed 1 test of 1.
  0.407927 not ok 1 - basic
  0.408038 #   Failed test 'basic'
  0.408041 #   at t/50proxy-forward-close-connection.t line 48.
  0.408077 failed to listen to 127.0.0.1:43397:Address already in use at t/Util.pm line 452.
  0.408612 # Tests were run but no plan was declared and done_testing() was not seen.
  0.408696 # Looks like your test exited with 111 just after 1.
  0.412929 Command exited with non-zero status 111
  0.412986 0.18user 0.04system 0:00.41elapsed 54%CPU (0avgtext+0avgdata 30080maxresident)k
  0.413034 0inputs+0outputs (0major+13163minor)pagefaults 0swaps
# t/50proxy-forward-close-connection.t failed
```
I read test code, some tests has not use Util::empty_ports even if it uses several ports listening. Thus,I consider that it failed cause it might have been using duplicate listening ports.

In this PR, I will fix such tests. If I completed fix all of these tests, I would change reviewable this.


kazuho (commented):
Great PR! It hadn't occurred to me the possibility of `empty_port` returning the same port number.

While I'm not fully certain if addressing that would reduce our failure rate (IMO failure rate is much higher than what we'd expect from such collision), I am inclined to adopt some kind of logic that prevents collision.

Regarding how, I am a bit concerned regarding lack of enforcement in the proposed approach. People can easily miss the fact that they have to sometime use `empty_ports` rather than `empty_port`.

Therefore, I wonder if we could define a wrapper of `Net::EmptyPort::empty_port` within `t::Util` that tracks the port numbers being returned to avoid collision? E.g.,

```perl
package t::Util;
...
my %empty_ports_returned;
sub empty_port {
    while (1) {
        my $p = Net::EmptyPort::port(@_);
        next if $empty_ports_returned{$p};
        $empty_ports_returned{$p} = 1;
        return $p;
    }
}
```

WDYT?

u5surf:
@kazuho 
Thank you for reviewing. And that' clever suggests! Certainly, I considered that we are not clear how to use new argment reserve_port.  In this case, wide scoped variables is not evil approach at all and not have to care more sentively cause it's just test code. 
I'll change this approach!

PR #2936, opened 2022-02-03
Dockerfile cleanup: unpin Test::TCP

miyagawa:
Test::TCP 2.21 has been widely available on CPAN and there shouldn't be a need to pin to a specific version on Git.


miyagawa:
by the way there's a few `sudo` commands inside the Dockerfile but these commands run as root anyway (at least before `USER ci` lines) so they're technically unnecessary.

gfx:
Sounds good to me. Also, +1 to remove the use of `sudo` in Dockerfiles. We need `sudo` command to be installed, however, because the tests require `sudo` to switch between a normal user and the superuser.

FWIW we need to push new images to DockerHub https://hub.docker.com/r/h2oserver/h2o-ci manually. Tell me your DockerHub account if you'll push images to DockerHub.

miyagawa:
OK, removed unnecessary `sudo` in ce299ac7381fa0d7363e587c63a6605578766b4d

gfx (approved):
LGTM.

I've updated the images on DockerHub and kicked CI. The result is all green.

miyagawa:
Thanks!

PR #2949, opened 2022-02-19
Fix output message which couldn't see occasionally.

u5surf:
`h2o server (pid:%d) is ready to serve requests with %zu threads`  sometimes couldn't appear in several tests.

e.g. 
```
# t/40http3-corrupted-scid-initial.t --------------------------------
  0.179642 spawning ./build/h2o... [INFO] raised RLIMIT_NOFILE to 1048576
  0.197682 session ticket secrets have been (re)loaded
  0.214420 fetch-ocsp-response (using OpenSSL 1.1.1  11 Sep 2018)
  0.214623 fetch-ocsp-response (using OpenSSL 1.1.1  11 Sep 2018)
  0.217543 failed to extract ocsp URI from /tmp/fYIFLVkU_K/cert.crt
  0.233533 failed to extract ocsp URI from /tmp/bXNX_9ZVyb/cert.crt
  0.233549 [OCSP Stapling] disabled for certificate file:examples/h2o/server.crt
  0.236492 [OCSP Stapling] disabled for certificate file:examples/h2o/server.crt
  0.268511 done
  1.402670 ok 1 - server is alive after getting an Initial with corrupted SCID (1)
  2.412682 ok 2 - server is alive after getting an Initial with corrupted SCID (2)
  2.421758 ok 3 - Number of connections is two
  3.449471 ok 4 - server is alive after getting an Initial with zero-length SCID (1)
  4.459632 ok 5 - server is alive after getting an Initial with zero-length SCID (2)
  4.468501 ok 6 - Number of connections is still two
  4.468576 killing ./build/h2o... received SIGTERM, gracefully shutting down
 14.580324 failed, sending SIGKILL... killed (got 9)
 14.580709 1..6
 14.585674 0.30user 0.05system 0:14.58elapsed 2%CPU (0avgtext+0avgdata 32480maxresident)k
 14.585695 4160inputs+0outputs (4major+21931minor)pagefaults 0swaps
# t/40http3-corrupted-scid-initial.t succeeded
```
https://github.com/h2o/h2o/blob/a429117babff09542d3517c4fa36c1ef769889c1/src/main.c#L3126-L3131

in case of `h2o_barrier_wait_pre_sync_point` seems not to return 1, it couldn't be seen them.

https://github.com/h2o/h2o/blob/a429117babff09542d3517c4fa36c1ef769889c1/lib/common/multithread.c#L308-L316

And if `barrier->_count == 0`, `h2o_barrier_wait_pre_sync_point` returns 1. 

OTOH, h2o_barrier_wait(h2o_barrier_wait_pre_sync_point) is called several points around this source code, e.g, t/40http3-corrupted-scid-initial.t is called it in which registers the session ticket in register_session_tickets. 

https://github.com/h2o/h2o/blob/a429117babff09542d3517c4fa36c1ef769889c1/src/ssl.c#L449-L459

If `barrier->_count` became 0 by h2o_barrier_wait here,  `barrier->_count` could not be 0 by `h2o_barrier_wait_pre_sync_point` which is called from main loop at the start up. Thus I consider that it could not seen that message.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -3127,8 +3127,10 @@ static void *run_loop(void *_thread_index)
         h2o_set_signal_handler(SIGTERM, on_sigterm_set_flag_notify_threads);
         if (conf.shutdown_requested)
             exit(0);
-        fprintf(stderr, "h2o server (pid:%d) is ready to serve requests with %zu threads\n", (int)getpid(), conf.thread_map.size);
     }
+    if (thread_index == 0)
+        fprintf(stderr, "h2o server (pid:%d) is ready to serve requests with %zu threads\n", (int)getpid(), conf.thread_map.size);
+
     h2o_barrier_wait_post_sync_point(&conf.startup_sync_barrier);
 
     /* the main loop */


kazuho (commented):
Thank you for the PR. Nice find, it is an issue that was missed in #2915.

I think the observation captures something more than a missing message. When the barrier in src/ssl.c is the last one to pass, not only the message is lost but the SITGERM handler is never registered.

I think we have to fix that as well.

Regarding how, the guarantee that `server_notifications` would be registered for all threads is met only when all threads call `h2o_barrier_wait_pre_sync_point`. Therefore, the proposed approach would not work for setting the signal handler at the correct moment.

I think we might want to:
* Partially revert #2915 that introduced `h2o_barrier_wait_pre_sync_point` and `h2o_barrier_wait_post_sync_point`.
* Instead, add a callback parameter to barrier, that would be called when the last thread enters the barrier.
* In src/main.c, register a callback that sets the sigterm handler and emits the message.

WDYT?

u5surf:
@kazuho 
Thanks reviewing. Although I'm not sure completely what can I do more, but I'm willing to try on this issue.
Of course I agree with the barrier in `src/ssl.c` also has to register SIGTERM handler.

> - Partially revert https://github.com/h2o/h2o/pull/2915 that introduced h2o_barrier_wait_pre_sync_point and h2o_barrier_wait_post_sync_point.
> - Instead, add a callback parameter to barrier, that would be called when the last thread enters the barrier.

After reverting https://github.com/h2o/h2o/pull/2915 partially, for example, we introduce the new callback function params  in `h2o_barrier_wait`  which fires when `barrier->_count == 0`.
```
-int h2o_barrier_wait(h2o_barrier_t *barrier)
+void h2o_barrier_wait(h2o_barrier_t *barrier, void (*cb) (void))
 {
-    int ret;
     pthread_mutex_lock(&barrier->_mutex);
     barrier->_count--;
     if (barrier->_count == 0) {
         pthread_cond_broadcast(&barrier->_cond);
-        ret = 1;
+        (*cb)();
     } else {
-        while (barrier->_count)
+        while (barrier->_count != 0)
             pthread_cond_wait(&barrier->_cond, &barrier->_mutex);
-        ret = 0;
     }
     pthread_mutex_unlock(&barrier->_mutex);
     /* This is needed to synchronize h2o_barrier_dispose with the exit of this function, so make sure that we can't destroy the
      * mutex or the condition before all threads have exited wait(). */
     __sync_sub_and_fetch(&barrier->_out_of_wait, 1);
-    return ret;
 }
```

> In src/main.c, register a callback that sets the sigterm handler and emits the message.

We could define following function which is to be the callback func params in h2o_barrier_wait.
And then it could be called both of main.c
```
+void h2o_set_sigterm_if_last_barrier(void)
+{
+    h2o_set_signal_handler(SIGTERM, on_sigterm_set_flag_notify_threads);
+    fprintf(stderr, "h2o server (pid:%d) is ready to serve requests with %zu threads\n", (int)getpid(), conf.thread_map.size);
+}
```
in run_loop, it just register a callback on your suggestions.
```
@@ -3121,17 +3128,10 @@ static void *run_loop(void *_thread_index)
     /* and start listening */

-    /* Wait for all threads to become ready but before letting any of them serve connections, swap the signal handler for graceful
-     * shutdown, check (and exit) if SIGTERM has been received already. */
-    if (h2o_barrier_wait_pre_sync_point(&conf.startup_sync_barrier)) {
-        h2o_set_signal_handler(SIGTERM, on_sigterm_set_flag_notify_threads);
-        if (conf.shutdown_requested)
-            exit(0);
-    }
-    if (thread_index == 0)
-        fprintf(stderr, "h2o server (pid:%d) is ready to serve requests with %zu threads\n", (int)getpid(), conf.thread_map.size);
-
-    h2o_barrier_wait_post_sync_point(&conf.startup_sync_barrier);
+    /* make sure all threads are initialized before starting to serve requests */
+    h2o_barrier_wait(&conf.startup_sync_barrier, h2o_set_sigterm_if_last_barrier);
+    if (conf.shutdown_requested)
+        exit(0);
```
Also as ssl.c like main.c
```
@@ -454,7 +454,7 @@ static void register_session_tickets(void (*cb)(void *), void *arg)
     pthread_rwlock_unlock(&session_tickets.rwlock);
     __sync_add_and_fetch(&session_tickets.generation, 1);
     if (session_tickets.barrier != NULL) {
-        h2o_barrier_wait(session_tickets.barrier);
+        h2o_barrier_wait(session_tickets.barrier, h2o_set_sigterm_if_last_barrier);
         session_tickets.barrier = NULL;
     }
 }
```
It seems to be good but ssl.c could not registered exit(0); correctly. 
Because `on_sigterm_set_flag_notify_threads` (in short `on_sigterm`) only can be swiched on main.c's `conf.shutdown_requested` flag, ssl.c could not see that flag.

We might just revert simply this diff.

https://github.com/h2o/h2o/pull/2915/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176L2496-L2500

u5surf:
@kazuho I ended up this PR. and reconsider ed about this issue by an another PR #2951

PR #2997, opened 2022-04-21
Fix up Prometeus by removing the requests JSON object

jlcox1970:
remove the h2o_requests response in the prometheus scrape output as the JSON object that is returned causes a parsing error


i110 (approved):
I'm not sure what kind of error you saw, but yeah requests data is hard to encode and not so useful as a prometheus stat, so 👍 for removing it.

PR #2998, opened 2022-04-21
Separate `startup_sync_barrier` into two barriers, the first only for run_loop threads.

jhatala:
The run_loop-thread-only barrier is named `startup_sync_barrier_init`.  It gates one-time initialization code for run_loop threads.  Only run_loop threads synchronize on this barrier.

The second barrier is named `startup_sync_barrier_post`.  It does not gate any one-time initialization code for run_loop threads.  It therefore does not matter whether the last thread to cross the barrier is a run_loop thread or not.

Before this PR, when the last thread to cross the barrier was the thread that runs `register_session_tickets`, then the initialization code for run_loop that the barrier was gating did not run.

This is an alternative to #2951.  Unlike #2951, this approach keeps the gated initialization code running on one of the run_loop threads.  Our internal additions to h2o rely on that.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -276,7 +276,8 @@ static struct {
         h2o_multithread_receiver_t memcached;
     } * threads;
     volatile sig_atomic_t shutdown_requested;
-    h2o_barrier_t startup_sync_barrier;
+    h2o_barrier_t startup_sync_barrier_init;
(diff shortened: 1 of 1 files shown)


jhatala:
Superseded by #2999, which is better,

PR #3020, opened 2022-05-20
nextbuf == NULL is not an error if there was no data to send

deweerdt:


kazuho (approved):
Thank you for the fix and the tests. To be merged once CI concludes.

PR #3021, opened 2022-05-26
when CONNECT method is used, pretend as if `priority: i` was sent

sleepybishop:
This PR cherry-picks tunnel priority fix from https://github.com/h2o/h2o/compare/kazuho/connect-fixes


kazuho (approved):
Thank you for the PR.

Yes, we need this, as we do not want tunnels getting blocked by other tunnels with smaller stream IDs, when the `priority` header field is missing.

To be merged once CI concludes.

PR #3030, opened 2022-06-03
[ci] take code coverage using LLVM source-based code coverage

gfx:
This PR adds a test scenario `coverage` to continuously take code coverage reports with LLVM souce-based code coverage in the continous mode.

I'll compare the baseline (75% in line-based coverage) with curl+http3 testing branch https://github.com/h2o/h2o/pull/2990, but after all it would be nice if we can take it with a code-coverage recording service such as [Codecov](https://about.codecov.io/). 

This PR also updates clang / LLVM to v14 because older LLVM could not have taken the code coverage very well.  

## References

* LLVM source-based code coverage https://clang.llvm.org/docs/SourceBasedCodeCoverage.html

## TODOs for the Next PRs

* Introduce a coverage traking service
* Merge the coverage stats in fuzzling tests


gfx:
As of [4ca544c]:(https://github.com/h2o/h2o/pull/3030/commits/4ca544c538b2eecb53c86ed41217198a231b33a5)

~~~
Filename                                         Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TOTAL                                              47143             15336    67.47%        2575               447    82.64%       45699             11331    75.21%       25276             10829    57.16%
~~~

gfx:
`llvm-cov -show-branches percent` shows branch conditions such as:

Screen Shot 2022-06-30 at 17 05 21

(the full command that I used to generate the above report is `llvm-cov show -instr-profile curl-with-h3.profdata -show-branches percent -format html ./b/h2o ./lib ./src > report.html`)

gfx:
FWIW there are TODOs before merging this PR:

* update the docker image `2004`
* revert [eb09864](https://github.com/h2o/h2o/pull/3030/commits/eb09864c87c2c7900ff2c57b67dbc4ff3aba457a)

t/90h2olog.t
```diff
@@ -8,6 +8,9 @@ use Test::More;
 use JSON;
 use t::Util;
 
+plan skip_all => "DTRACE_TESTS=0 is set to skip this file"
+    if defined($ENV{DTRACE_TESTS}) && !$ENV{DTRACE_TESTS};
```
kazuho, line 12:
```suggestion
    unless $ENV{DTRACE_TESTS};
```
Maybe this is sufficient?

kazuho (commented):
Thank you for working on this. Looks good.

Discussed offline with @gfx that we'd better use ubuntu2204 image added in #3053.

t/90h2olog.t
gfx: I'd like to run this test file if `DTRACE_TESTS` is not set. That is, my intension is:

* `DTRACE_TESTS=1` to force dtrace tests
* `DTRACE_TESTS=0` to skip dtrace tests
* no `DTRACE_TESTS` specified  to run dtrace tests if prerequisites are met

gfx:
Changed to use the ubuntu2204 image + the `llvm` package!

gfx:
`CI / OpenSSL 3.0` is likely to fail with the following logs:

```
 26.571886     # Subtest: TCP
 26.661185         ok 1 - req: HTTP/1
 27.662352         not ok 2 - nothing has been logged
 27.662723         #   Failed test 'nothing has been logged'
 27.662741         #   at t/90h2olog-selective-tracing.t line 141.
 27.662915         #          got: '{"type":"socket-write","tid":46929,"seq":1,"sock":139752062398016,"bufs":139752326506608,"bufcnt":1,"time":1657866880448}
 27.662922         # {"type":"socket-write-tls-record","tid":46929,"seq":2,"sock":139752062398016,"write-size":9,"bytes-buffered":31,"time":1657866880448}
 27.662925         # {"type":"socket-writev","tid":46929,"seq":3,"sock":139752062398016,"ret":31,"time":1657866880448}
 27.662928         # {"type":"socket-writev","tid":46929,"seq":4,"sock":139752062398016,"ret":24,"time":1657866880449}
 27.662930         # '
 27.662933         #     expected: ''
 27.663384         1..2
 27.663512         # Looks like you failed 1 test of 2.
 27.663842     not ok 1 - TCP
```

but I cannot reproduce it with my local env (Ubuntu 22.04 + OpenSSL 3.0.2).

kazuho:
@gfx Could the culprit be us running multiple h2olog tests at once?

I ask this because in one of the failures, I see something like:
```
2022-07-15T07:14:14.2214982Z  14.028266 0inputs+0outputs (0major+8454minor)pagefaults 0swaps
2022-07-15T07:14:14.2215400Z [92m# t/90h2olog.t succeeded
(snip)
022-07-15T07:15:08.3437434Z  74.451135 0inputs+0outputs (0major+8481minor)pagefaults 0swaps
2022-07-15T07:15:08.3437900Z [91m# t/90h2olog-selective-tracing.t failed
2022-07-15T07:15:08.3438404Z [0m--------------------------------------------------------
2022-07-15T07:15:08.3438787Z [31m# Result: Fail
```

First column (e.g., `2022-07-15T07:14:14.2214982Z`) is when the line was passed to GitHub Actions. Second column (e.g., `14.028266`) is how long it has taken since the test process was launched by `run_tests`.

By subtracting latter from the former, we can say that t/90h2olog.t started at 07:14.00.193 and ended at 07:15:08.343, t/90h2olog-selective-tracitng.t started at 07:13:53.892 and ended at 07:15:08.343.

If that is the case, you might be able to replicate the problem by using `run_tests` to run two tests (or more) at once.

t/90h2olog.t
kazuho: I'm not sure if that's consistent with how we `DTRACE_TESTS`. Can I ask why you think something like that is necessary?

Rereading existing code, I have several feelings:
* It is a mistake to refer to `DTRACE_TESTS` in h2olog.t - it should be calling `check_dtrace_availability` instead.
* The rule around `DTRACE_TESTS` is probably unneed. In `check_dtrace_availability`, probably all we need to do is check prerequisites, including the version check of `bpftrace`. If `bpftrace` is too old, we should skip tests. Otherwise, we should run them. There is no need to have a flag that allows a user to specify whether some test should be run.

WDYT?

kazuho: Discussed offline with @gfx, agreed that t/90h2olog.t should simply depend on `check_dtrace_availability`.

We will keep `check_dtrace_availability` as is, for the time being. The complexity of running dtrace tests is not only on the linux side; on macOSX, it requires a boot parameter change too.

gfx: I could remove the changes in test files after all. `check_dtrace_availability()` has a bit different purpose, which checks tracer tools such as `dtrace(1)` or `bpftrace(1)`. 

I'd like to refactor around them (e.g. coming up with a new utility function that are shared in all the dtrace / h2olog tests), but will do in a separate PR.

gfx:
I believe it's ready to merge!

kazuho:
Looks great! Thank you for the work.

PR #3035, opened 2022-06-15
http2client: properly send request body with 0-bytes

miyagawa:
When client sends a POST request with `content-length: 0`, h2o reverse proxy skips sending the data frame for the request because the size is 0. This is an incorrect behavior and could make the upstream h2 server get stalled waiting for the data frame to arrive. This patch fixes it so that it will send the data frame for the 0-byte body.

Added a unit test for the behavior, confirmed it failed without the patch in 5d0ee4cbdabacf4745064f4dc049e9fc7a04ccb1, and is fixed with the patch.


kazuho (approved):
Thank you for the fix. Inclusion of a test that confirms the existence of bug and the fix is reassuring.

PR #3045, opened 2022-07-06
misc http2client fixes

miyagawa:
- fix assertion failures when linking same timeout multiple times
- fix post failures when sending post body larger than 16kb


miyagawa:
closing in favor of #3047

PR #3049, opened 2022-07-08
Add h2olog probes for CONNECT and CONNECT-UDP tunnels.

jhatala:
New probes:
- `connect-tcp-attempt` - An attempt to connect on the backend-side of a CONNECT tunnel.
- `connect-tcp-write` - A write to the TCP connection on the backend-side of a CONNECT tunnel.
- `connect-tcp-write-error` - A write error on the TCP connection on the backend-side of a CONNECT tunnel.
- `connect-tcp-read` - A read from the TCP connection on the backend-side of a CONNECT tunnel.
- `connect-tcp-read-error` - A read error on the TCP connection on the backend-side of a CONNECT tunnel.
- `connect-udp-attempt` - An attempt to connect on the backend-side of a CONNECT-UDP tunnel.
- `connect-udp-write` - A write to the UDP connection on the backend-side of a CONNECT-UDP tunnel.
- `connect-udp-read` - A read from the UDP connection on the backend-side of a CONNECT-UDP tunnel.
- `connect-error` - Error trying to establish a CONNECT or CONNECT-UDP tunnel.
- `connect-io-timeout` - Idle timeout on a CONNECT or CONNECT-UDP tunnel.
- `connect-dispose` - Done handling a CONNECT or CONNECT-UDP request.


lib/handler/connect.c
```diff
@@ -123,6 +124,13 @@ struct st_connect_generator_t {
 
 static void record_error(struct st_connect_generator_t *self, const char *error_type, const char *details, const char *rcode)
 {
+    if (H2O_CONN_IS_PROBED(CONNECT_ERROR, self->src_req->conn)) {
+        char msg1[64], msg2[64]; // STR_LEN in src/h2olog/misc/gen_raw_tracer.py
```
gfx, line 128: IIRC h2olog discards a string that is longer than STR_LEN, so we don't need to adjust it here.

Also, as you might know, the BPF program in h2olog does not compile if an event object is too large so STR_LEN is restricted to a small value, but in this case, it seems that we don't need to contact error_type and rcode. 

So 
`H2O_PROBE_CONN(CONNECT_ERROR, self->src_req->conn, self->src_req->req_index, error_type, rcode, details);` might be enough. What do you think?

gfx: Ah, it's the case where the event object is too long!:

> error: /virtual/main.c:1806:13: in function trace_quicly__stream_send i32 (%struct.pt_regs*): Looks like the BPF stack limit of 512 bytes is exceeded. Please move large on stack variables into BPF per-cpu array map.

Please ignore the above comment.

lib/probes_.h
```diff
@@ -95,6 +95,7 @@ __attribute__((noinline)) static void h2o_probe_response_header(h2o_req_t *req,
 
 static inline void h2o_probe_log_request(h2o_req_t *req, uint64_t req_index)
 {
+    req->req_index = req_index;
```
kazuho, line 98: This looks like a hack, in terms of both assigning the value here and having duplicate state (i.e., having `req_index` here and stream ID in h2 / h3 layer).

How about:
* adding a mandatory callback to `h2o_conn_callbacks_t` that returns the request index given `h2o_req_t *` as an argument, and
* creating `H2O_PROBE_REQUEST` macro like `H2O_PROBE_CONN`, that takes `h2o_req_t *` as the first argument and spits connection ID and request ID to eBPF?

lib/handler/connect.c
```diff
@@ -533,8 +553,14 @@ static void tcp_on_read(h2o_socket_t *_sock, const char *err)
 
     if (err == NULL) {
         h2o_iovec_t vec = h2o_iovec_init(self->sock->input->bytes, self->sock->input->size);
+        H2O_PROBE_CONN(CONNECT_TCP_READ, self->src_req->conn, self->src_req->req_index, vec.len);
         h2o_send(self->src_req, &vec, 1, H2O_SEND_STATE_IN_PROGRESS);
     } else {
+        if (err == h2o_socket_error_closed) {
+            H2O_PROBE_CONN(CONNECT_TCP_READ_CLOSED, self->src_req->conn, self->src_req->req_index);
+        } else {
+            H2O_PROBE_CONN(CONNECT_TCP_READ_ERROR, self->src_req->conn, self->src_req->req_index, err);
```
kazuho, line 562: Am I correct to understand that the intent for having two probes that show closure is to have the possibility of just monitoring abnormal behavior, rather than monitoring all close events?

If so that's fine (but otherwise I would prefer merging the two).

kazuho (commented):
Thank you for the PR. I like this, it looks good modulo the points below.

lib/handler/connect.c
kazuho: Yeah I think it might be possible to reduce stack pressure pretty easily.

At the moment, `trace_quicly__stream_send` looks like this:
```c
// quicly:stream_send
int trace_quicly__stream_send(struct pt_regs *ctx) {
  const void *buf = NULL;
  struct h2olog_event_t event = { .id = H2OLOG_EVENT_ID_QUICLY_STREAM_SEND, .tid = (uint32_t)bpf_get_current_pid_tgid(), };

  // struct st_quicly_conn_t * conn
  uint8_t conn[sizeof_st_quicly_conn_t] = {};
  bpf_usdt_readarg(1, ctx, &buf);
  bpf_probe_read(&conn, sizeof_st_quicly_conn_t, buf);
  event.stream_send.conn_master_id = get_st_quicly_conn_t__master_id(conn);
  // int64_t at
  bpf_usdt_readarg(2, ctx, &event.stream_send.at);
  // struct st_quicly_stream_t * stream
  uint8_t stream[sizeof_st_quicly_stream_t] = {};
  bpf_usdt_readarg(3, ctx, &buf);
  bpf_probe_read(&stream, sizeof_st_quicly_stream_t, buf);
  event.stream_send.stream_stream_id = get_st_quicly_stream_t__stream_id(stream);
  // uint64_t off
  bpf_usdt_readarg(4, ctx, &event.stream_send.off);
  // size_t len
  bpf_usdt_readarg(5, ctx, &event.stream_send.len);
  // int is_fin
  bpf_usdt_readarg(6, ctx, &event.stream_send.is_fin);

  if (events.perf_submit(ctx, &event, sizeof(event)) != 0)
    bpf_trace_printk("failed to perf_submit in trace_quicly__stream_send\n");

  return 0;
}
```

The pressure comes from `uint8_t conn[sizeof_st_quicly_conn_t]` and `uint8_t stream[sizeof_st_quicly_stream_t]` sharing the same stack.

I think (or would hope) that if we could use blocks (`{}`) for each argument so that `conn` and `stream` do not coexist at one time, then we can free 100 bytes of stack space.

gfx: > I think (or would hope) that if we could use blocks ({}) for each argument so that conn and stream do not coexist at one time, then we can free 100 bytes of stack space.


Confirmed https://github.com/h2o/h2o/pull/3051 makes this PR compile with three string params. Let's remove the hack to merge strings.

lib/handler/connect.c
jhatala: I was thinking that it is useful to know whether it was the client or the origin who decided to close the connection, that is why there is write-closed probe and a read-closed probe.

I do not have a good reason for separating read-closed from read-error.  I will see if I can make it work sensibly with only one probe here.

lib/probes_.h
jhatala: That is a great idea, thank you!

lib/handler/connect.c
```diff
@@ -520,6 +536,10 @@ static int tcp_write(void *_self, int is_end_stream)
     if (self->sock != NULL && !h2o_socket_is_writing(self->sock))
         tcp_do_write(self);
 
+    if (is_end_stream) {
+        H2O_PROBE_CONN(CONNECT_TCP_WRITE_CLOSED, self->src_req->conn, self->src_req->req_index);
```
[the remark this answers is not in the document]
jhatala: I was thinking that it is useful to know whether it was the client or the origin who decided to close the connection, that is why there is write-closed probe and a read-closed probe.

But maybe we want to have just one probe in `close_socket` and report in a parameter what caused the close.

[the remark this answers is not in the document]
kazuho: > I was thinking that it is useful to know whether it was the client or the origin who decided to close the connection, that is why there is a write-closed probe and a read-closed probe.

I think that’s a reasonable argument.

My arguments are:
* The name of the probe being proposed might be confusing, because it follows the same naming convention as others, but unlike others, monitors the interaction with the connection handler (lib/httpX) rather than the socket.
* If our intent is to monitor interaction with the connection handler, maybe what we need is a set of probes for monitoring all reads and write events that happen with the connection handler, rather than just the close event on one direction.
* The interface between the generator (e.g., lib/handler/connect.c) and connection handler is not unique to lib/handler/connect.c. Probably, what we want to have is wrapper functions around that interface, and within those wrapper functions, define necessary probes. Then, we will have a generator-agnostic way of monitoring the interactions.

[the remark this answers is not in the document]
jhatala: If we have the `connect-dispose` probe and the `connect-tcp-read-error` probe reports the origin closing the connection, then looking at the trace we can easily determine which side initiated the connection closure:

```json
{"type":"connect-tcp-read","tid":31102,"seq":14,"conn-id":1,"req-id":1,"bytes-len":24,"time":1658536106802}
{"type":"connect-tcp-read-error","tid":31102,"seq":15,"conn-id":1,"req-id":1,"err":"socket closed by peer","time":1658536106802}
{"type":"connect-dispose","tid":31102,"seq":16,"conn-id":1,"req-id":1,"time":1658536106802}
```
vs
```json

```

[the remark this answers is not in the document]
jhatala: That makes sense, thank you!

If we have the `connect-dispose` probe and the `connect-tcp-read-error` probe reports the origin closing the connection, then looking at the trace we can easily determine which side initiated the connection closure.

The origin-side closed first:
```json
["... earlier log entries removed ..."]
{"type":"connect-tcp-read","tid":31102,"seq":14,"conn-id":1,"req-id":1,"bytes-len":24,"time":1658536106802}
{"type":"connect-tcp-read-error","tid":31102,"seq":15,"conn-id":1,"req-id":1,"err":"socket closed by peer","time":1658536106802}
{"type":"connect-dispose","tid":31102,"seq":16,"conn-id":1,"req-id":1,"time":1658536106802}
```

The client-side closed first:
```json
["... earlier log entries removed ..."]
{"type":"connect-tcp-write","tid":31103,"seq":31,"conn-id":2,"req-id":1,"bytes-len":24,"time":1658536296758}
{"type":"connect-dispose","tid":31103,"seq":32,"conn-id":2,"req-id":1,"time":1658536296758}
```

We do not need `connect-tcp-read-closed` and `connect-tcp-write-closed` for that.

lib/handler/connect.c
```diff
@@ -254,6 +262,9 @@ static void close_readwrite(struct st_connect_generator_t *self)
 static void on_io_timeout(h2o_timer_t *timer)
 {
     struct st_connect_generator_t *self = H2O_STRUCT_FROM_MEMBER(struct st_connect_generator_t, timeout, timer);
+    if (self->sock != NULL) {
```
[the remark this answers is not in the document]
jhatala: Interestingly, without the if, I see the timeout handler run twice for an HTTP/3 client connection:

```
{"type":"connect-io-timeout","tid":44371,"seq":41,"conn-id":6,"req-id":0,"time":1658539411374}
{"type":"connect-io-timeout","tid":44371,"seq":42,"conn-id":6,"req-id":0,"time":1658539411374}
```

And zero times when the client connection is HTTP/2 with `h2o-httpclient`, but that is a separate mystery to look into.

lib/handler/connect.c
```diff
@@ -123,6 +124,13 @@ struct st_connect_generator_t {
 
 static void record_error(struct st_connect_generator_t *self, const char *error_type, const char *details, const char *rcode)
 {
+    if (H2O_CONN_IS_PROBED(CONNECT_ERROR, self->src_req->conn)) {
+        char msg1[64], msg2[64]; // STR_LEN in src/h2olog/misc/gen_raw_tracer.py
+        snprintf(msg1, sizeof(msg1), "%s%s%s", error_type, (rcode != NULL ? " " : ""), (rcode != NULL ? rcode : ""));
+        snprintf(msg2, sizeof(msg2), "%s", (details != NULL ? details : "(null)"));
+        H2O_PROBE_CONN(CONNECT_ERROR, self->src_req->conn, self->src_req->req_index, msg1, msg2);
```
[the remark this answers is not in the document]
jhatala: Done in b5c36230f.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: `if` removed in 21c8f1dfa.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: Probes `connect-tcp-read-closed` and `connect-tcp-write-closed` removed in 21c8f1dfa.

lib/handler/connect.c
```diff
@@ -823,6 +854,7 @@ static void on_stop(h2o_generator_t *_self, h2o_req_t *req)
 static void on_generator_dispose(void *_self)
 {
     struct st_connect_generator_t *self = _self;
+    H2O_PROBE_CONN(CONNECT_DONE, self->src_req->conn, self->src_req->req_index);
```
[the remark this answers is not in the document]
jhatala: Done in 21c8f1dfa.

lib/probes_.h
jhatala: Done in 21c8f1dfa.

lib/handler/connect.c
jhatala: Probe `connect-tcp-read-closed` removed in 21c8f1dfa.

lib/handler/connect.c
jhatala: Done in b5c36230f.

jhatala:
In the test run "CI / OpenSSL 3.0 (Ubuntu 22.04 / DTrace)", `t/50fastcgi.t` failed, but the following makes me suspect that it is a flake caused by a port collision:

```
2022-07-23T03:17:01.4935409Z   4.451631 spawning /home/ci/h2o... [/tmp/qnl7_bwuoK:12] in command listen, failed to listen to TCP port 0.0.0.0:37591: Address already in use
```

lib/probes_.h
jhatala: Done in cd07df4.

lib/handler/connect.c
jhatala: Done in d489a36.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: Done in d489a36.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: `if` removed in cd07df4.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: The two probes removed in cd07df4.

lib/handler/connect.c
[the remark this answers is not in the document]
jhatala: Rename done in cd07df4.

lib/handler/connect.c
jhatala: It is now one probe since cd07df4.

kazuho:
Thank you the changes. Merged with slight changes in #3068.

jhatala:
Nice, thank you!

PR #3078, opened 2022-08-03
Make sure limit is not too short

hfujita:
This is a follow up of h2o/h2o#2635, to ensure `limit` is long enough to include "...".

### DIFF
--- a/lib/core/request.c
+++ b/lib/core/request.c
@@ -697,6 +697,7 @@ static size_t append_with_limit(char *dst, h2o_iovec_t input, size_t limit)
         memcpy(dst, input.base, input.len);
         return input.len;
     } else {
+        assert(limit >= 3);
         memcpy(dst, input.base, (limit - 3));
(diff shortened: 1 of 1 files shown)


hfujita:
Maybe I'm just too paranoid 😄

lib/core/request.c
```diff
@@ -697,6 +697,7 @@ static size_t append_with_limit(char *dst, h2o_iovec_t input, size_t limit)
         memcpy(dst, input.base, input.len);
         return input.len;
     } else {
+        assert(limit >= 3);
```
gfx, line 700: I think we needn't declare this kind of assertions, as we don't put `assert(dst != NULL)`, `assert(input.base != NULL)` and so on. If you were worried about the contract, it would be good enough to declare these contracts in a comment.

hfujita: Okay makes sense. I vaguely remember that I had a similar discussion before. Withdrawing this PR then.

PR #3083, opened 2022-08-05
Use the `closefrom` in `libbsd` if available.

jhatala:
This helps with startup time when `stace`-ing `h2o` on a Linux machine with a large number of available (but not open) file descriptors.

On Linux, `libbsd` implements `closefrom` using the list of file descriptors in `/proc/self/fd/`.

In order to confirm that the closing of extra file descriptors still works, I started `h2o` under `strace` with `777</dev/null` and looked for the file descriptor number `777` being closed.

Before this change:
```
14502 prlimit64(0, RLIMIT_NOFILE, NULL,  <unfinished ...>
14502 <... prlimit64 resumed>0x7ffcbd4f41c0) = 0
14502 close(1048575)                    = -1 EBADF (Bad file descriptor)
14502 close(1048574)                    = -1 EBADF (Bad file descriptor)
[... a million lines...]
14502 close(9)                          = -1 EBADF (Bad file descriptor)
14502 close(7)                          = -1 EBADF (Bad file descriptor)
14502 close(5)                          = 0
14502 close(4)                          = 0
14502 close(3)                          = 0
```

After this change:
```
23226 close(3 <unfinished ...>
23226 <... close resumed>)              = 0
23226 close(4 <unfinished ...>
23226 <... close resumed>)              = 0
23226 close(5)                          = 0
23226 close(7)                          = -1 EBADF (Bad file descriptor)
23226 openat(AT_FDCWD, 0x7f245b798000, O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 3
23226 fstat(3, 0x7fffc9581c60)          = 0
23226 getdents64(3, 0x5603a22a7c10, 32768) = 216
23226 getdents64(3, /* 0 entries */, 32768) = 0
23226 close(777)                        = 0
23226 close(3)                          = 0

```


jhatala:
`t/50mruby-redis.t` failed in the run "CI / OpenSSL 3.0 (Ubuntu 22.04 / DTrace) (pull_request)" seemingly due to a port collision:

```
2022-08-05T07:55:11.7738003Z   0.468218 spawning /home/ci/h2o... [/tmp/EE3etJDOKp:17] in command listen, failed to listen to TCP port 0.0.0.0:45047: Address already in use
```

kazuho (commented):
@jhatala Thank you for the PR.

Does https://github.com/h2o/neverbleed/pull/39 work as a fix to your problem?

* The changes have to go to h2o/neverbleed.
* neverbleed does not have a build script - we'd rather want to avoid doing feature detection (like looking for libbsd) there.
* I think we have to have the loop-based fallback in h2o/neverbleed, for operating systems such as macOSX.

Based on these observations, I wonder if we could address the issue just by adding an invocation to `close_range` before falling back to the loop-based logic.

jhatala:
>Does https://github.com/h2o/neverbleed/pull/39 work as a fix to your problem?

Yes, I think that it does.  Thank you!

PR #3093, opened 2022-08-22
Do not interact with `self` if it was disposed

deweerdt:


gfx:
cf. https://github.com/h2o/h2o/pull/3085

PR #3116, opened 2022-09-21
check setsockopt return values

gfx:
It must not raise fatal errors because a request is being accepted at that time, so extracting these parts in a method and sharing it with the QUIC setting up routine do not make sense.

discussions:
* https://github.com/h2o/h2o/pull/3003
* https://github.com/h2o/h2o/pull/3095


kazuho (approved):
Thank you for the changes.

This PR looks fine. Please merge once the CI succeeds.

While it might be a good idea to rate-limit these errors being reported (see how we restrict the number of epoll misfires in `evloop_do_proceed` function of lib/common/socket/evloop/epoll.c.h), we can do that as a next step, probably by extending `h2o_perror`.

PR #3141, opened 2022-10-24
[WIP] h2olog v2: removing BCC dependency

gfx:


gfx:
Discussed with @kazuho elsewhere and we've decided not to implement a h2olog-specific custom protocol. We'll implement HTTP for h2olog v2 instead. Closing this and will make another PR.

PR #3152, opened 2022-11-04
Return newly created DSR sender correctly

hfujita:
This is a follow up to #3124, which failed to return a newly created DSR sender object.


hfujita:
This will be applied to #2415

PR #3158, opened 2022-12-03
CONNECT, CONNECT-UDP: report the destination IP as param next-hop in proxy-status on error and on success.

jhatala:


lib/handler/connect.c
```diff
@@ -56,6 +56,7 @@ struct st_connect_generator_t {
         struct st_server_address_t list[MAX_ADDRESSES_PER_FAMILY * 2];
         size_t size;
         size_t used;
+        const struct st_server_address_t *trying;
```
kazuho, line 59: Is this field ever going to be different from `server_addresses.list[server_addresses.used]`?

lib/handler/connect.c
```diff
@@ -145,13 +163,45 @@ static void record_error(struct st_connect_generator_t *self, const char *error_
...
+    if (addr_str_len == SIZE_MAX)
+        return;
+
+    h2o_iovec_t parts[] = {
+        get_proxy_status_identity(self),
+        h2o_iovec_init(H2O_STRLIT("; next-hop=")),
+        h2o_iovec_init(addr_str, addr_str_len),
+    };
+    const size_t nparts = sizeof(parts) / sizeof(parts[0]);
+    h2o_iovec_t hval = h2o_concat_list(&self->src_req->pool, parts, nparts);
```
kazuho, line 199:
```suggestion
    h2o_iovec_t hval = h2o_concat(&self->src_req->pool, get_proxy_status_identity(self), h2o_iovec_init(H2O_STRLIT("; next-hop=")), h2o_iovec_init(addr_str, addr_str_len));
```

Contrary to the logic we have in `record_error`, the list being supplied here is always the same, therefore we can use `h2o_concat`.

As an alternative, I think there is a possibility of creating a unified function dedicated to building the header field, regardless of the result being success or failure. Doing so would reduce code duplication. But I think that's a matter of taste, I'm fine either ways.

kazuho (commented):
Thank you for the PR. I like the direction.

lib/handler/connect.c
jhatala: Yes, that is simpler, thank you.

I will give the unified function a try, it should work, because the success response simply does not have the `error`, `rcode` and `details` parameteers.  Thank you for the suggestion!

jhatala: Done in 332703ce5.

The call to `h2o_socket_getnumerichost` is outside of `add_proxy_status_header` because that allows the `h2o_req_log_error` call in `record_error` to log the IP address without doing the conversion twice.

What do you think?

lib/handler/connect.c
kazuho: Regarding how it works, I think you observation is correct.

Regarding if it is a behavior that we can depend on, I think it is, due to following reasons. Maybe what we need is a doc-comment around the structure regarding how the data structure is used:
* Of the entries in `st_connect_generator_t::server_addresses`, elements [0, used) contain addresses to which we have attempted to (or currently under an attempt to) connect.
* [used, size) contain addresses to which we have not.
* Considering that we need to at least have a list of addresses to which we might try to connect in the future, it makes sense to have this type of data structure.
* Due to us dividing an array into two logical lists, when trying a new address, it makes sense to pick the one at the border of the two lists.
* When trying to connect to a new address, we want to pick an address in the other address family if such an address exists. That's the logic around `pick_and_swap`.

FWIW, I might agree that It might seem fragile to rely on `used` to determine the address to which we have been currently trying to connect to. But if that is the concern, I might argue that the proposed approach (of retaining a pointer to the last address being used) is also fragile. Both of the approaches assume that the addresses that we have tried are not moved, and that only one address is being tried at a time. If we are to implement Happy Eyeballs in full spec, we would not have the second guarantee.

Therefore, if we want something solid, I think we should retain a copy of the address to which we are trying to connect to, rather than a pointer to an entity containing such address.

That said, I am not sure if we need such complexity when existing structure is capable of telling what the address is. Hence my original comment.

jhatala: An alternative implementation that removes the new member is in #3163.

`get_dest_addr_str` optionally returns the value so that the probe `connect_success` probe can return it.  If we do not want the detail in the probe (or we do not even want the `connect_success` probe, then `get_dest_addr_str` can be simplified and it will be an even smaller change.

Please let me know if think that that is a better trade-off to do it that way than to add `trying` to the state.

jhatala: Done in f47fd98b1.

`get_dest_addr_str` optionally returns the `struct st_server_address_t` pointer so that the probe `connect_success` can log it.  (If we do not want the new probe, then `get_dest_addr_str` can be simplified and the PR will be an even smaller change.)

kazuho (commented):
Thank you for the changes. Looks good, I only have very minor suggestions. How about the following?

lib/handler/connect.c
```diff
@@ -116,40 +116,91 @@ struct st_connect_generator_t {
...
+        identity = h2o_iovec_init(H2O_STRLIT("h2o"));
+    return identity;
+}
+
+static h2o_iovec_t get_dest_addr_str(struct st_connect_generator_t *self, char *dest_addr_strbuf,
+                                     const struct st_server_address_t **addr_out)
+{
+    const struct st_server_address_t *addr = NULL;
+    h2o_iovec_t dest_addr_str = h2o_iovec_init(NULL, 0);
+    if (self->server_addresses.used > 0 && self->server_addresses.used <= self->server_addresses.size) {
```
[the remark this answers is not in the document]
jhatala: I think that we need something for when the function is called from `record_error` because that can happen before there are any addresses.  Would you be fine with just `if (self->server_addresses.used > 0)`?  That should detect the case when there are no IP addresses yet correctly.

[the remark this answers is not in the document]
kazuho: Ah, thanks for pointing out the error case. I see it now.

lib/handler/connect.c
```diff
@@ -116,40 +116,91 @@ struct st_connect_generator_t {
...
+        addr = &self->server_addresses.list[self->server_addresses.used - 1];
+        size_t len = h2o_socket_getnumerichost(addr->sa, addr->salen, dest_addr_strbuf);
+        if (len != SIZE_MAX) {
+            dest_addr_str = h2o_iovec_init(dest_addr_strbuf, len);
+        }
+    }
+    if (addr_out != NULL) {
+        *addr_out = addr;
+    }
```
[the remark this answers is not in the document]
jhatala: Done in bac0104df.

kazuho (approved):
Thank you for the changes. Looks great.

PR #3316, opened 2023-12-01
Expect: 100-continue

i110:
This PR does 2 different things related to 100 continue:

1. h2 and h3 protocol handlers immediately send `100 continue` response when they find `expect: 100-continue` header, just like h1
2. add `proxy.expect-100-continue` config knob which let http1client wait `100 continue` response before sending request body

TODOs:
- http2client and http3client should also wait 100 response
- add timeout configuration like below so as not to have httpclient wait indefinitely (another PR?)
```
 proxy.expect-100-continue:
    timeout: 1000
```


lib/common/http1client.c
```diff
@@ -789,23 +817,54 @@ static void start_request(struct st_h2o_http1client_t *client, h2o_iovec_t metho
     client->_is_chunked = *props->chunked;
     client->_method_is_head = h2o_memis(method.base, method.len, H2O_STRLIT("HEAD"));
 
+    if (client->proceed_req != NULL || body.len != 0) {
```
kazuho, line 820: Do we want to indicate use of `expect: 100-continue` by adding a property `h2o_httpclient_properties_t`, rather than using a header field?

I ask this because we handle the expect header as if it was a hop-by-hop header in the server-side. The idea is that it might make sense to be consistent on both the server- and client-side.

lib/common/http1client.c
```diff
@@ -608,7 +634,9 @@ static void req_body_send_complete(h2o_socket_t *sock, const char *err)
 
     int is_end_stream = client->body_buf.is_end_stream;
 
-    call_proceed_req(client, NULL);
+    if (client->proceed_req != NULL) {
+        call_proceed_req(client, NULL);
+    }
```
kazuho, line 639: This looks like a change to the state machine. Could you please elaborate how you found this to be an issue and why you think this is the correct way to fix?

lib/common/http1client.c
```diff
@@ -595,6 +610,17 @@ static void on_whole_request_sent(h2o_socket_t *sock, const char *err)
     }
 }
 
+static void on_header_sent(h2o_socket_t *sock, const char *err)
```
kazuho, line 613: Maybe rename the function to `on_headers_sent_wait_100` or something?

Current name might be confusing in sense that it sounds as if this function is always called after the request headers have been sent.

lib/common/http1client.c
```diff
@@ -595,6 +610,17 @@ static void on_whole_request_sent(h2o_socket_t *sock, const char *err)
...
+{
+    struct st_h2o_http1client_t *client = sock->data;
+
+    h2o_timer_unlink(&client->super._timeout);
```
kazuho, line 617: Are we fine with just disabling the timer?

I get that we do not need to ream it when an error has occurred, but unless that is the case, I think we should be arming either an I/O timeout or a first-byte timeout?

lib/core/proxy.c
```diff
@@ -225,6 +227,10 @@ static void build_request(h2o_req_t *req, h2o_iovec_t *method, h2o_url_t *url, h
                 } else if (token == H2O_TOKEN_EARLY_DATA) {
                     found_early_data = 1;
                     goto AddHeader;
+                } else if (token == H2O_TOKEN_EXPECT) {
+                    /* this never happens as protocol handlers should have consumed expect header */
+                    h2o_error_printf("[WARN] expect header unexpectedly found on http %x\n", req->version);
+                    continue;
```
kazuho, line 233: I'm not sure if we need to be worried specifically about the expect header getting through, even though we do not have these fail checks for other headers; e.g., Connection ,Upgrade, ...

lib/core/proxy.c
```diff
@@ -733,7 +743,11 @@ static int on_1xx(h2o_httpclient_t *client, int version, int status, h2o_iovec_t
             h2o_push_path_in_link_header(self->src_req, headers[i].value.base, headers[i].value.len);
     }
 
-    if (status != 101) {
+    if (status == 101) {
+        /* switching protocol */
+    } else if (status == 100) {
+        /* we don't need to forward 100 since protocol handlers have already done */
```
kazuho, line 749: Considering that lib/common/http1client.c consumes and reacts to 100, do we need to have the signal propagated to proxy.c?

kazuho (commented):
Thank you for your efforts! I've not read the test yet but please see my comments below.

lib/common/http1client.c
i110: This is actually not changing state machine, but is just reusing the existing code. Before this PR, `req_body_send` was only used when req streaming is enabled, but this PR utilizes that function also in the case of non-streaming case. In that case the client doesn't have `proceed_req`, so we need this check to support both cases.

lib/common/http1client.c
```diff
@@ -465,7 +478,9 @@ static void on_head(h2o_socket_t *sock, const char *err)
         }
         h2o_buffer_consume(&client->sock->input, rlen);
         if (client->sock->input->size == 0) {
-            h2o_timer_link(client->super.ctx->loop, client->super.ctx->io_timeout, &client->super._timeout);
+            if (! h2o_timer_is_linked(&client->super._timeout)) {
+                h2o_timer_link(client->super.ctx->loop, client->super.ctx->io_timeout, &client->super._timeout);
+            }
```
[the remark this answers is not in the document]
i110: Thank you, but I just applied `clang-format` at https://github.com/h2o/h2o/pull/3316/commits/03f8d32e29a69f3ed3813031ecded75b4e1e9a91 since your suggestion seems to be removing `!` :)

Or did you actually want to remove braces?

lib/common/http1client.c
```diff
@@ -457,6 +458,18 @@ static void on_head(h2o_socket_t *sock, const char *err)
         if (!(100 <= http_status && http_status <= 199 && http_status != 101))
             break;
 
+        if (http_status == 100 && client->_expect_100_continue) {
+            /* start sending request body */
+            if (client->proceed_req != NULL) {
+                if (client->body_buf.buf->size != 0) {
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/237219c9e6fd8a2e6ab90be61125b280ba6e68b8

lib/common/http1client.c
```diff
@@ -457,6 +458,18 @@ static void on_head(h2o_socket_t *sock, const char *err)
         if (!(100 <= http_status && http_status <= 199 && http_status != 101))
             break;
 
+        if (http_status == 100 && client->_expect_100_continue) {
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/fa2b8996341ae450a63c95896f593d6b93a29cd8

Talking about the name of the field, I feel `_expect_100_continue` is reasonable enough because it means just expecting 100, and turning it off obviously means `it is not expecting 100`, but I don't have strong opinion :)

lib/common/http1client.c
i110: Basically agreed. I remember I've been struggled with this design, and there must be some reason I finally chose this way, but now I feel it's more straightforward to use h2o_httpclient_properties_t.

TBD

lib/common/http1client.c
i110: https://github.com/h2o/h2o/pull/3316/commits/3c7b85016a17ccee04ef5d0b76f80e92b02225e2

lib/core/proxy.c
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/4f22cbc3c60c5a0363925db7026ab3ee574d693e

lib/core/proxy.c
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/0ce40027978ba0b383a8095c44ad55f4e0e1b98c

lib/common/http1client.c
i110: Partially addressed in https://github.com/h2o/h2o/pull/3316/commits/5770af1e4f79482ae58e5eacdb43eba3dc5d6da5
I came up with one TODO I added as a comment. If the response status of stream is `HEADER`, or it's not `HEADER` but the received response headers include `expect` header, there must be no issues. Otherwise, i.e. http1client already receives response header but it didn't contain `expect` header, waiting 100 means stall. Should we let it stall or return error to the client? Or start sending req body regardless of the absence of expect header? I appreciate your suggestion.

lib/common/http1client.c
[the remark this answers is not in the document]
kazuho: Of course you are correct!

lib/common/http1client.c
kazuho: Very good question! When receiving responses to Expect: 100-continue, I think what we want to do is:
* if the response is 100, start sending the request body
* if the response is 1xx other than 100, continue waiting for 100, while forwarding 1xx
* if the response is final (i.e, other than 100 or 1xx), start sending the request body.

Regarding the last bullet point, note that there's nothing that prohibits servers start sending a final response before receiving a complete request. RFC 9112 (HTTP/1.1) might be silent about this, but [RFC 9113 section 8.1](https://httpwg.org/specs/rfc9113.html#HttpFraming) explicitly states that "a server can send a complete response prior to the client sending an entire request if the response does not depend on any portion of the request that has not been sent and received."

Under such situation, while we might not need to send the entire request, I think it might make sense for us to try sending the entire request until the server closes the connection, recalling the design choices that we have made in the past for handling these early responses in our server-side code.

lib/common/http1client.c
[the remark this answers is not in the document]
kazuho: SG. But separately, please see my comment regarding if this logic should be specific to 100 and nothing else.

lib/common/http1client.c
i110: Thank you for the suggestion. https://github.com/h2o/h2o/pull/3316/commits/4dade945ae45cdd215772b4e107622fe0c4e621f makes h2o start sending request body when it receives final response.

lib/common/http1client.c
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/ba9dd9696d9a9cca49604d69d46e37f1d4ebbd26

kazuho (commented):
Thank you for the changes!

lib/core/proxy.c
i110: Ah sorry for the lack of explanation, yeah, considering your comments I thought that consuming here is consistent and can simplify the things. Thanks!

lib/common/http1client.c
```diff
@@ -457,15 +458,20 @@ static void on_head(h2o_socket_t *sock, const char *err)
         if (!(100 <= http_status && http_status <= 199 && http_status != 101))
             break;
 
-        if (client->super.informational_cb != NULL &&
-            client->super.informational_cb(&client->super, version, http_status, h2o_iovec_init(msg, msg_len), headers,
-                                           num_headers) != 0) {
+        if (http_status == 100 && client->_expect_100_continue) {
+            client->_expect_100_continue = 0;
+            req_body_send(client);
```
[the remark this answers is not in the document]
i110: I'd think this is a topic of software layers. I felt the current way, which is, `http1client itself can forward expect: 100-continue request and 100 response, while it's configurable by use_hcd flag, but proxy absorbs it` is one of a reasonable designs. What do you think?

lib/common/http1client.c
```diff
@@ -550,6 +556,14 @@ static void on_head(h2o_socket_t *sock, const char *err)
     h2o_buffer_consume(&sock->input, rlen);
     client->_socket_bytes_processed = client->sock->bytes_read - client->sock->input->size;
 
+    if (client->_expect_100_continue) {
+        /* we have suspended request body, let's start sending it.
+         * see: https://github.com/h2o/h2o/pull/3316#discussion_r1456859634
+         */
+        client->_expect_100_continue = 0;
+        req_body_send(client);
+    }
```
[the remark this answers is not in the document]
i110: Could you a bit more elaborate how we can merge those logics? IMHO `the one above that handles 100` runs upon 100 response, while we want to run this code upon final responses. I'm not sure how we can merge them

lib/common/http1client.c
```diff
@@ -789,30 +827,53 @@ static void start_request(struct st_h2o_http1client_t *client, h2o_iovec_t metho
...
+            size_t bytes_written;
+            reqbufcnt += req_body_send_prepare(client, reqbufs + reqbufcnt, &bytes_written);
+            client->super.bytes_written.body = bytes_written;
+            h2o_socket_write(client->sock, reqbufs, reqbufcnt, req_body_send_complete);
+        }
+    } else if (body.len != 0) {
         assert(!client->_is_chunked);
-        if (body.len != 0) {
+        if (client->_expect_100_continue) {
+            h2o_buffer_init(&client->body_buf.buf, &h2o_socket_buffer_prototype);
```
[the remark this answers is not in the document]
i110: Thank you, you're right. Addressed in https://github.com/h2o/h2o/pull/3316/commits/84eaa0ddd48eea1d3966627af983de85796f40da

lib/common/http1client.c
```diff
@@ -758,8 +791,13 @@ static h2o_iovec_t build_request(struct st_h2o_http1client_t *client, h2o_iovec_
...
         APPEND_HEADER(&u);
-    } else if (connection.base != NULL) {
-        h2o_header_t h = {&H2O_TOKEN_CONNECTION->buf, NULL, connection};
+    } else if (props->connection_header->base != NULL) {
+        h2o_header_t h = {&H2O_TOKEN_CONNECTION->buf, NULL, *props->connection_header};
+        APPEND_HEADER(&h);
+    }
+
+    if (props->expect_100_continue) {
+        h2o_header_t h = {&H2O_TOKEN_EXPECT->buf, NULL, h2o_iovec_init(H2O_STRLIT("100-continue"))};
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/618b091338b62663fce89aa85f58e9af7085f06f

lib/common/http1client.c
```diff
@@ -789,30 +827,53 @@ static void start_request(struct st_h2o_http1client_t *client, h2o_iovec_t metho
...
 
     client->_is_chunked = *props->chunked;
     client->_method_is_head = h2o_memis(method.base, method.len, H2O_STRLIT("HEAD"));
 
+    if (props->expect_100_continue) {
+        if (client->proceed_req == NULL && body.len == 0) {
+            /* no body */
+            on_whole_request_sent(client->sock, h2o_httpclient_error_internal);
+            return;
+        }
```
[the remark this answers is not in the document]
i110: ditto

include/h2o.h
```diff
@@ -1109,6 +1109,10 @@ typedef struct st_h2o_req_overrides_t {
      * whether the proxied request should preserve host
      */
     unsigned proxy_preserve_host : 1;
+    /**
+     * whether the proxied request send expect: 100-continue and wait 100 response before sending request body
+     */
+    unsigned proxy_expect_100_continue : 1;
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/198140c6bf30f30481b57f73988119c4a33d356d

include/h2o.h
```diff
@@ -2209,6 +2213,7 @@ typedef struct st_h2o_proxy_config_vars_t {
         uint64_t connection_attempt_delay;
     } happy_eyeballs;
     unsigned preserve_host : 1;
+    unsigned expect_100_continue : 1;
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/198140c6bf30f30481b57f73988119c4a33d356d

include/h2o/httpclient.h
```diff
@@ -67,6 +67,11 @@ typedef struct st_h2o_httpclient_properties_t {
      * defaults to false
      */
     unsigned prefer_pipe_reader : 1;
+    /**
+     * When the value is 1, httpclient sends 'expect: 100-continue' header and suspend sending request body
+     * until it sees 100 continue response
+     */
+    unsigned expect_100_continue : 1;
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/198140c6bf30f30481b57f73988119c4a33d356d

include/h2o/httpclient.h
```diff
@@ -67,6 +67,11 @@ typedef struct st_h2o_httpclient_properties_t {
      * defaults to false
      */
     unsigned prefer_pipe_reader : 1;
+    /**
+     * When the value is 1, httpclient sends 'expect: 100-continue' header and suspend sending request body
+     * until it sees 100 continue response
```
kazuho, line 72:
```suggestion
     * When the value is 1, httpclient sends 'expect: 100-continue' header and suspends sending request body
     * until it sees 100-continue response
```

lib/common/http1client.c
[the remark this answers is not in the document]
kazuho: I think we still have a problem here; to me it seems that this if block rejects sending non-streamed HTTP requests with zero-byte of body.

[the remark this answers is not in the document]
i110: Thank you so much! I think the right thing to do here is just do nothing if the req has no body, instead of erroring out. Addressed in https://github.com/h2o/h2o/pull/3316/commits/64aa6217b8cc1d2d64a5f172aeae1b481754788a

lib/common/http1client.c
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3316/commits/102b6d292da60ffb459daf1cea70ef7798dac1d5

lib/common/http1client.c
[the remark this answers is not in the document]
kazuho: Agreed elsewhere to forward 100 always at the httpclient-level. In terms of h2o server, lib/core/proxy.c absorbs 100-continue unconditionally, in fear of unexpected 100s causing clients to fail.

lib/common/http1client.c
```diff
@@ -454,18 +455,32 @@ static void on_head(h2o_socket_t *sock, const char *err)
             headers[i].flags = (h2o_header_flags_t){0};
         }
 
-        if (!(100 <= http_status && http_status <= 199 && http_status != 101))
+        if (http_status == 101) {
+            assert(!client->_use_expect);
```
i110, line 459: @kazuho One question: this means that a malicious upstream server can intentionally crash h2o by sending a 101 response to h2o that sent an 'expect' header. Is this safe?

kazuho: Of course you are correct. Fixed in 5f695e4 and 29f807e.

kazuho:
@i110 Thank you for all your efforts, I think this PR is ready modulo one problem: H3 is broken:
```
# Subtest: h3
    # Subtest: streaming
        ok 1
        1..1
    ok 1 - streaming
    # Subtest: non-streaming
Assertion failed: (stream->state == H2O_HTTP3_SERVER_STREAM_STATE_RECV_BODY_BEFORE_BLOCK || stream->state == H2O_HTTP3_SERVER_STREAM_STATE_SEND_HEADERS || stream->state == H2O_HTTP3_SERVER_STREAM_STATE_SEND_BODY), function on_send_shift, file /Users/kazuho/mydev/h2o/lib/http3/server.c, line 900.
received fatal signal 6
[13392] 0   h2o                                 0x0000000103fea9c8 on_sigfatal + 88
[13392] 1   libsystem_platform.dylib            0x00007ff8042025ed _sigtramp + 29
[13392] 2   ???                                 0x0000000000000087 0x0 + 135
[13392] 3   libsystem_c.dylib                   0x00007ff8040fbb45 abort + 123
[13392] 4   libsystem_c.dylib                   0x00007ff8040fae5e err + 0
[13392] 5   h2o                                 0x0000000104105563 on_send_shift.cold.1 + 35
[13392] 6   h2o                                 0x0000000103f6e8f8 on_send_shift + 504
[13392] 7   h2o                                 0x0000000103eea600 on_ack_stream_ack_one + 272
[13392] 8   h2o                                 0x0000000103ef8841 handle_ack_frame + 2193
[13392] 9   h2o                                 0x0000000103edfde3 handle_payload + 211
[13392] 10  h2o                                 0x0000000103ee09be quicly_receive + 2558
[13392] 11  h2o                                 0x0000000103f682c2 process_packets + 2466
[13392] 12  h2o                                 0x0000000103f678e3 h2o_quic_read_socket + 1491
[13392] 13  h2o                                 0x0000000103f1694c run_socket + 380
[13392] 14  h2o                                 0x0000000103f12c36 h2o_evloop_run + 2022
[13392] 15  h2o                                 0x0000000103fe2ac3 run_loop + 1587
[13392] 16  h2o                                 0x0000000103fe1845 main + 5381
[13392] 17  dyld                                0x00007ff803e7b41f start + 1903
        not ok 1
        #   Failed test at t/40expect.t line 60.
        #                   'HTTP/3 100 
        # server: h2o/2.3.0-DEV@29f807e10
        # 
        # '
        #     doesn't match '(?^si:HTTP/[0-9.]+ 100\s.*HTTP/[0-9.]+ 200\s)'
        1..1
        # Looks like you failed 1 test of 1.
    not ok 2 - non-streaming
    #   Failed test 'non-streaming'
    #   at t/40expect.t line 61.
    1..2
    # Looks like you failed 1 test of 2.
not ok 3 - h3
#   Failed test 'h3'
#   at t/40expect.t line 63.
1..3
```

You might have missed this because we do not yet have curl with http3 support in CI. We need to resume work on #2990, but would it be possible for you to look into this crash before #2990 gets merged?

FWIW I used curl with http3 support that I have my hand outside of CI, and ran the test directly.

i110:
hmm thank you for noticing this, but unfortunately I cannot repro with curl supporting HTTP3. Will give a detailed look later

i110:
@kazuho I managed to reproduce it, and believe https://github.com/h2o/h2o/pull/3316/commits/f6be12245f4cb3d1f6980b2c19178976024f5860 is the fix!

PR #3341, opened 2024-02-01
add missing http2-max-streams srcdoc

i110:
https://github.com/h2o/h2o/pull/3337 seems to be missing srcdoc for `http2-max-streams` directive. That causes `t/60find-doc.t` to fail after I regenerated the doc.


antoniofastly (approved):
Thanks!

PR #3358, opened 2024-02-24
Fix proxy_idle_time to exclude client latency

jcbran:
This change measures proxy_idle_time as the time from when we create the backend client until we receive the first byte from the backend.

Prior to this patch we measured the time from when we received the first byte from the client until we received the first byte from the backend which inadvertently included network and client latency resulting in inflated proxy_idle_times.


deweerdt:
@jcbran thank you for the PR, there seems to be an issue with the tests, would you mind taking a look?

PR #3405, opened 2024-07-31
Reverse Tunnel over HTTP

i110:
This is the *server*-side implementation of [Reverse Tunnel over HTTP](https://www.ietf.org/archive/id/draft-kazuho-httpbis-reverse-tunnel-00.html).

```
listen:
    type: reverse
    client: https://127.0.0.1:36256/.well-known/reverse/tcp/example.com/8080
    ssl:
        key-file: /path/to/key
        certificate-file: /path/to/cert
        verify-peer: ON
```

As you see in the above conf, this also supports mtls feature for upgrading request. It's one of a valid and useful authentication schemes as [the spec](https://www.ietf.org/archive/id/draft-kazuho-httpbis-reverse-tunnel-00.html#section-4) describes.

Now this only supports HTTP/1 (i.e. http1client) as the protocol for initiating tunnel connection, but it'd be sufficient now considering that the *clients* can send h2 requests upon that connection.

Also this supports mtls 

## TODOs:
  - write integration tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,6 +396,7 @@ SET(LIB_SOURCE_FILES
     lib/common/memory.c
     lib/common/multithread.c
     lib/common/redis.c
+    lib/common/reverse_tunnel.c
     lib/common/serverutil.c
     lib/common/socket.c
     lib/common/socketpool.c
--- a/include/h2o/reverse_tunnel.h
+++ b/include/h2o/reverse_tunnel.h
@@ -4,6 +4,7 @@ struct st_h2o_accept_ctx_t;
 
 typedef struct st_h2o_reverse_config_t {
     uint64_t reconnect_interval;
+    SSL_CTX *ssl_ctx;
     void (*setup_socket)(h2o_socket_t *sock, void *data);
 } h2o_reverse_config_t;
 
--- a/include/h2o/reverse_tunnel.h
+++ b/include/h2o/reverse_tunnel.h
@@ -1,0 +1,26 @@
+#include "h2o/httpclient.h"
+
+struct st_h2o_accept_ctx_t;
+
+typedef struct st_h2o_reverse_config_t {
+    uint64_t reconnect_interval;
+    void (*setup_socket)(h2o_socket_t *sock, void *data);
+} h2o_reverse_config_t;
+
+typedef struct st_h2o_reverse_ctx_t {
+    h2o_reverse_config_t config;
+    h2o_url_t *client;
+    struct st_h2o_accept_ctx_t *accept_ctx;
+    h2o_timer_t reconnect_timer;
+    void *data;
+    struct {
+        h2o_httpclient_t *client;
+        h2o_httpclient_ctx_t ctx;
+        h2o_httpclient_connection_pool_t connpool;
+        h2o_socketpool_t sockpool;
+    } httpclient;
+    h2o_mem_pool_t pool;
+} h2o_reverse_ctx_t;
+
+void h2o_reverse_start_listening(h2o_reverse_ctx_t *reverse);
+void h2o_reverse_init(h2o_reverse_ctx_t *reverse, h2o_url_t *client, struct st_h2o_accept_ctx_t *accept_ctx, h2o_reverse_config_t config, void *data);
--- a/lib/common/http1client.c
+++ b/lib/common/http1client.c
@@ -560,6 +560,11 @@ static void on_head(h2o_socket_t *sock, const char *err)
     if (client->state.res == STREAM_STATE_CLOSED) {
         close_response(client);
         return;
+    } else if (client->super._cb.on_body == h2o_httpclient_socket_stealed) {
+        h2o_buffer_consume(&sock->input, rlen);
+        client->sock = NULL;
+        close_client(client);
+        return;
     } else if (client->super._cb.on_body == NULL) {
         client->_do_keepalive = 0;
         close_client(client);
--- a/lib/common/httpclient.c
+++ b/lib/common/httpclient.c
@@ -302,3 +302,9 @@ void h2o_httpclient_set_conn_properties_of_socket(h2o_socket_t *sock, h2o_httpcl
     properties->ssl.cipher_bits = h2o_socket_get_ssl_cipher_bits(sock);
     properties->sock = sock;
 }
+
+static int socket_stealed(h2o_httpclient_t *client, const char *errstr, h2o_header_t *trailers, size_t num_trailers)
+{
+    assert(!"this callback is solery for signaling purpose and should never be called");
+}
+const h2o_httpclient_body_cb h2o_httpclient_socket_stealed = socket_stealed;
--- a/lib/common/reverse_tunnel.c
+++ b/lib/common/reverse_tunnel.c
@@ -1,0 +1,195 @@
+/*
+ * Copyright (c) 2024 Ichito Nagata, Fastly, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include "h2o/reverse_tunnel.h"
+#include "h2o.h"
+
+static inline void schedule_reconnect(h2o_reverse_ctx_t *reverse)
+{
+    h2o_timer_link(reverse->accept_ctx->ctx->loop, reverse->config.reconnect_interval, &reverse->reconnect_timer);
+}
+
+struct on_reverse_close_data_t {
+    void (*orig_cb)(void *data);
+    void *orig_data;
+    h2o_reverse_ctx_t *reverse;
+};
+
+void on_reverse_close(void *_data) {
+    struct on_reverse_close_data_t *data = (void *)_data;
+    data->orig_cb(data->orig_data);
+
+    h2o_reverse_ctx_t *reverse = data->reverse;
+    reverse->httpclient.client = NULL;
+    h2o_mem_clear_pool(&reverse->pool);
+
+    schedule_reconnect(reverse);
+}
+
+static void on_reverse_reconnect_timeout(h2o_timer_t *timer)
+{
+    h2o_reverse_ctx_t *reverse = H2O_STRUCT_FROM_MEMBER(h2o_reverse_ctx_t, reconnect_timer, timer);
+    h2o_reverse_start_listening(reverse);
+}
+
+static void on_reverse_read(h2o_socket_t *sock, const char *err)
+{
+    h2o_reverse_ctx_t *reverse = (void *)sock->data;
+
+    h2o_socket_read_stop(sock);
+    if (err != NULL) {
+        h2o_error_printf("unexpected read error in on_reverse_rea: %s\n", err);
+        return;
+    }
+
+    h2o_accept(reverse->accept_ctx, sock);
+}
+
+static h2o_httpclient_body_cb on_reverse_head(h2o_httpclient_t *client, const char *errstr, h2o_httpclient_on_head_t *args)
+{
+    h2o_reverse_ctx_t *reverse = (void *)client->data;
+
+    if (errstr != NULL) {
+        h2o_error_printf("received error in on_reverse_head: %s\n", errstr);
+        schedule_reconnect(reverse);
+        return NULL;
+    }
+
+    if (args->status != 101) {
+        h2o_error_printf("received unexpected status in on_reverse_head: %u\n", args->status);
+        schedule_reconnect(reverse);
+        return NULL;
+    }
+
+
+    // replace sock's on_close data with our own to retry on close
+    h2o_httpclient_conn_properties_t conn_props;
+    client->get_conn_properties(client, &conn_props);
+
+    struct on_reverse_close_data_t *data = h2o_mem_alloc_pool(&reverse->pool, *data, sizeof(*data));
+    data->reverse = reverse;
+    data->orig_cb = conn_props.sock->on_close.cb;
+    data->orig_data = conn_props.sock->on_close.data;
+    conn_props.sock->on_close.cb = on_reverse_close;
+    conn_props.sock->on_close.data = data;
+
+    conn_props.sock->data = reverse;
+
+    if (reverse->config.setup_socket != NULL)
+        reverse->config.setup_socket(conn_props.sock, reverse->data);
+
+    h2o_socket_read_stop(conn_props.sock);
+    h2o_socket_read_start(conn_props.sock, on_reverse_read);
+
+    return h2o_httpclient_socket_stealed;
+}
+
+// this callback is actually never called, but needed just to signal that streaming mode is enabled
+static void reverse_proceed_request(h2o_httpclient_t *client, const char *errstr)
+{
+    if (errstr != NULL)
+        h2o_error_printf("reverse_proceed_request failed: %s\n", errstr);
+}
+
+
+static h2o_httpclient_head_cb on_reverse_connect(h2o_httpclient_t *client, const char *errstr, h2o_iovec_t *method, h2o_url_t *url,
+                                         const h2o_header_t **headers, size_t *num_headers, h2o_iovec_t *body,
+                                         h2o_httpclient_proceed_req_cb *proceed_req_cb, h2o_httpclient_properties_t *props,
+                                         h2o_url_t *origin)
+{
+    h2o_reverse_ctx_t *reverse = (void *)client->data;
+
+    if (errstr != NULL) {
+        h2o_error_printf("error in on_reverse_connect: %s\n", errstr);
+        schedule_reconnect(reverse);
+        return NULL;
+    }
+
+    // h2o_hostconf_t *hostconf = conf.listeners[reverse->listener->listener_index]->hosts[0];
+
+    // char portbuf[sizeof(H2O_UINT16_LONGEST_STR)];
+    // size_t portlen = sprintf(portbuf, "%u", hostconf->authority.port);
+
+    // setup request
+    *method = h2o_iovec_init(H2O_STRLIT("GET"));
+    *url = *reverse->client;
+
+    // *url = *origin;
+    // url->path = h2o_concat(&reverse->pool,
+    //     h2o_iovec_init(H2O_STRLIT("/.well-known/reverse/tcp/")),
+    //     hostconf->authority.host,
+    //     h2o_iovec_init(H2O_STRLIT("/")),
+    //     h2o_iovec_init(portbuf, portlen)
+    // );
+
+    h2o_headers_t headers_vec = (h2o_headers_t){};
+    h2o_add_header_by_str(&reverse->pool, &headers_vec, H2O_STRLIT("ALPN"), 0, NULL, H2O_STRLIT("http%2F1.1"));
+    *headers = headers_vec.entries;
+    *num_headers = headers_vec.size;
+    *body = h2o_iovec_init(NULL, 0);
+    *proceed_req_cb = reverse_proceed_request;
+
+    return on_reverse_head;
+}
+
+void h2o_reverse_init(h2o_reverse_ctx_t *reverse, h2o_url_t *client, h2o_accept_ctx_t *accept_ctx, h2o_reverse_config_t config, void *data)
+{
+    reverse->client = client;
+    reverse->config = config;
+    reverse->accept_ctx = accept_ctx;
+    h2o_timer_init(&reverse->reconnect_timer, on_reverse_reconnect_timeout);
+    reverse->data = data;
+
+    reverse->httpclient.ctx = (h2o_httpclient_ctx_t){
+        .loop = accept_ctx->ctx->loop,
+        .getaddr_receiver = &accept_ctx->ctx->receivers.hostinfo_getaddr,
+        .io_timeout = 10000, // FIXME
+        .connect_timeout = 10000, // FIXME
+        .first_byte_timeout = 10000,
+        .keepalive_timeout = 10000,
+        .max_buffer_size = SIZE_MAX,
+        .protocol_selector = {.ratio = { .http2 = 0, .http3 = 0}}, // now only supports h1
+    };
+    reverse->httpclient.connpool = (h2o_httpclient_connection_pool_t){
+        .socketpool = &reverse->httpclient.sockpool,
+    };
+
+    h2o_socketpool_target_t *target = h2o_socketpool_create_target(client, NULL);
+    h2o_socketpool_init_specific(&reverse->httpclient.sockpool, SIZE_MAX, &target, 1, NULL);
+    h2o_socketpool_set_timeout(&reverse->httpclient.sockpool, UINT64_MAX);
+    h2o_socketpool_register_loop(&reverse->httpclient.sockpool, accept_ctx->ctx->loop);
+
+    SSL_CTX *ssl_ctx = SSL_CTX_new(SSLv23_client_method());
+    h2o_socketpool_set_ssl_ctx(&reverse->httpclient.sockpool, ssl_ctx);
+    SSL_CTX_free(ssl_ctx);
+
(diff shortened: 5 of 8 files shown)


CMakeLists.txt
```diff
@@ -396,6 +396,7 @@ SET(LIB_SOURCE_FILES
     lib/common/memory.c
     lib/common/multithread.c
     lib/common/redis.c
+    lib/common/reverse_tunnel.c
```
kazuho, line 399: The implementation depends on h2o.h, so it cannot go into lib/common. lib/core will be a better location.

Though, it might just be sufficient to add code to lib/core/util.c, considering that other logic that setup up server-side HTTP (proxy protocol handler and TLS handshake) exists there.

include/h2o/reverse_tunnel.h
```diff
@@ -0,0 +1,27 @@
...
+        h2o_httpclient_t *client;
+        h2o_httpclient_ctx_t ctx;
+        h2o_httpclient_connection_pool_t connpool;
+        h2o_socketpool_t sockpool;
+    } httpclient;
+    h2o_mem_pool_t pool;
+} h2o_reverse_ctx_t;
+
+void h2o_reverse_start_listening(h2o_reverse_ctx_t *reverse);
+void h2o_reverse_init(h2o_reverse_ctx_t *reverse, h2o_url_t *client, struct st_h2o_accept_ctx_t *accept_ctx, h2o_reverse_config_t config, void *data);
```
kazuho, line 27: Is there a reason to have two structures and two functions?

Maybe it would be sufficient to just have one that is per-thread.

lib/common/httpclient.c
```diff
@@ -302,3 +302,9 @@ void h2o_httpclient_set_conn_properties_of_socket(h2o_socket_t *sock, h2o_httpcl
     properties->ssl.cipher_bits = h2o_socket_get_ssl_cipher_bits(sock);
     properties->sock = sock;
 }
+
+static int socket_stealed(h2o_httpclient_t *client, const char *errstr, h2o_header_t *trailers, size_t num_trailers)
+{
+    assert(!"this callback is solery for signaling purpose and should never be called");
+}
+const h2o_httpclient_body_cb h2o_httpclient_socket_stealed = socket_stealed;
```
kazuho, line 310: Past participle of "steal" is "stolen", though I think we can call the function "h2o_httpclient_steal_socket".

Also, there is no need to define a static function and to have a global that takes the address. All we need is an external function with the correct name, e.g.,:
```suggestion
int h2o_httpclient_steal_socket(h2o_httpclient_t *client, const char *errstr, h2o_header_t *trailers, size_t num_trailers)
{
    assert(!"this callback is solely for signaling purpose and should never be called");
}
```

lib/common/socketpool.c
```diff
@@ -334,7 +334,8 @@ static void try_connect(h2o_socketpool_connect_request_t *req)
     switch (target->type) {
     case H2O_SOCKETPOOL_TYPE_NAMED:
         /* resolve the name, and connect */
-        req->getaddr_req = h2o_hostinfo_getaddr(req->getaddr_receiver, target->url.host, target->peer.named_serv, AF_UNSPEC,
+        req->getaddr_req = h2o_hostinfo_getaddr(req->getaddr_receiver, target->url.host, target->peer.named_serv, AF_INET,
+        // req->getaddr_req = h2o_hostinfo_getaddr(req->getaddr_receiver, target->url.host, target->peer.named_serv, AF_UNSPEC,
```
kazuho, line 338: I believe these changes are unnecessary?

src/main.c
```diff
@@ -2666,14 +2735,22 @@ static int on_config_listen_element(h2o_configurator_command_t *cmd, h2o_configu
             return -1;
         if (rcvbuf_node != NULL && h2o_configurator_scanf(cmd, *rcvbuf_node, "%u", &stream_rcvbuf) != 0)
             return -1;
+        if (reconnect_interval_node != NULL && h2o_configurator_scanf(cmd, *reconnect_interval_node, "%lu", &reconnect_interval) != 0)
```
kazuho, line 2738: Scanf designator for uint64_t is `"%" SCNu64`.

src/main.c
```diff
@@ -2666,14 +2735,22 @@ static int on_config_listen_element(h2o_configurator_command_t *cmd, h2o_configu
...
+            return -1;
+        if (connections_per_thread_node != NULL && h2o_configurator_scanf(cmd, *connections_per_thread_node, "%lu", &connections_per_thread) != 0)
+            return -1;
```
kazuho, line 2741: Here also? Though I think we'd be fine using `int` instead (see max-connections).

src/main.c
```diff
@@ -2666,14 +2735,22 @@ static int on_config_listen_element(h2o_configurator_command_t *cmd, h2o_configu
...
     } break;
     default:
         h2o_configurator_errprintf(cmd, node, "value must be a string or a mapping (with keys: `port` and optionally `host`)");
         return -1;
     }
 
-    if (strcmp(type, "unix") == 0) {
+    if (strcmp(type, "reverse") != 0 && servname == NULL) {
+        h2o_configurator_errprintf(cmd, *initcwnd_node, "cannot find mandatory attribute: port");
```
kazuho, line 2749:
```suggestion
        h2o_configurator_errprintf(cmd, node, "type is not `reverse` and `port` is not specified");
```
2nd argument should point to the mapping within which `port` must be defined.

src/main.c
```diff
@@ -1882,13 +1904,18 @@ static int listener_setup_ssl(h2o_configurator_command_t *cmd, h2o_configurator_
...
                                            &ocsp_update_interval_node, &ocsp_max_failures_node, &dh_file, &cipher_preference_node,
-                                           &neverbleed_node, &http2_origin_frame_node, &client_ca_file, &ech_node) != 0)
+                                           &neverbleed_node, &http2_origin_frame_node, &client_ca_file, &ech_node, &verify_peer_node) != 0)
             return -1;
         if (identity_node != NULL) {
+            if (is_reverse_listener(listener)) {
+                h2o_configurator_errprintf(cmd, *identity_node,
+                                           "identity directive is not supported in reverse listener");
+                return -1;
+            }
```
kazuho, line 1918: I think we can negotiate the type of client certificates, but we do not need to do that at this stage.

src/main.c
```diff
@@ -222,6 +223,14 @@ struct listener_config_t {
          */
         struct listener_config_t *sibling;
     } quic;
+
+    struct {
+        h2o_url_t client;
```
kazuho, line 228:
```suggestion
        h2o_url_t url;
```
Maybe. "client" is confusing in sense that it does not imply what property of the client the variable is. In contrast, there is only one "url" in the context of reverse HTTP establishment.

src/main.c
```diff
@@ -222,6 +223,14 @@ struct listener_config_t {
...
+        uint64_t reconnect_interval;
+        uint64_t connections_per_thread;
+        SSL_CTX *ssl_ctx;
```
kazuho, line 231: As `h2o_socketpool_t` is the structure that retains the SSL_CTX used by httpclient, it might make sense to have `h2o_socketpool_t` here instead.

src/main.c
```diff
@@ -4147,18 +4254,31 @@ H2O_NORETURN static void *run_loop(void *_thread_index)
...
                                                {&conf.threads[thread_index].ctx, listener_config->hosts, NULL, NULL,
                                                 listener_config->proxy_protocol, &conf.threads[thread_index].memcached}};
-        if (listener_config->ssl.size != 0) {
+        if (listener_config->ssl.size != 0 && !is_reverse_listener(listener_config)) {
             listeners[i].accept_ctx.ssl_ctx = listener_config->ssl.entries[0]->identities[0].ossl;
             listeners[i].accept_ctx.http2_origin_frame = listener_config->ssl.entries[0]->http2_origin_frame;
         }
-        listeners[i].sock = h2o_evloop_socket_create(conf.threads[thread_index].ctx.loop, fd, H2O_SOCKET_FLAG_DONT_READ);
-        listeners[i].sock->data = listeners + i;
+        if (is_reverse_listener(listener_config)) {
```
kazuho, line 4264: See my comment on reverse_tunnel.h. I do not think we need init and start_listening, or the need to retain the information in `listener_config`.

We can just setup the reverse listener and start listening after the ordinary listeners start accepting.

kazuho (commented):
Thank you for the PR.

The code looks mostly good, I think we are on the right path.

lib/common/socketpool.c
i110: Oh I mistakenly committed this change, this is due to some error (I don't remember) when ipv6 address was resolved. Will revert this back.

src/main.c
i110: But here we want have this ssl_ctx shared between threads, while socketpools are dedicated to each thread?

src/main.c
i110: I basically agree but would you mind clarifying what this means?

> start listening after the ordinary listeners start accepting.

Is it a requirement to start listening after other listeners start accepting? If so why?

src/main.c
kazuho: Socket pools are shared among threads.

src/main.c
kazuho: For ordinary listeners, we do the following:
1. setup the per-thread state for accepting connections in all the threads (this might hard fail)
2. setup graceful shutdown logic in thread zero
3. start accepting connections

But in the case of reverse http, there is no need to have separate steps (1 and 3), because the reverse channel for "accepting HTTP connections" is opened and reopened after the server starts running. Failures to setup the reverse channel is expected, and we will be treating that as a softfail, trying to reconnect periodically.

That's the reason everything can be done in step 3, and that we do not need init and start_listening as separate functions is reverse_channel.h.

CMakeLists.txt
i110: Done in https://github.com/h2o/h2o/pull/3405/commits/b75b2029350ff7af85286e6b4dd729d74baf1930

src/main.c
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/04307140c0b03eff2983fa9fbb6e2e9f1157388a.

> See https://github.com/h2o/h2o/pull/2913 regarding why we do have these steps.

#2913 and #2915 are a mix of several issues and fixes, and just by glancing them I couldn't find an answer to the question of why the initialization and the start of the accept process need to be separated. Appreciate it if you could explain it to me when you have time.

lib/common/httpclient.c
i110: Renamed the function in https://github.com/h2o/h2o/pull/3405/commits/d59ab4ab39d73ebe37956a0115477d54d097cf9d.
OTOH I'd think that the current way of using `extern const` prevents from repeating body_cb's arguments in httpclient.h, and it looks like a little clearer that it's just a constant to signal something, not the actual routine which would be invoked. What do you think?

src/main.c
i110: Renamed in https://github.com/h2o/h2o/pull/3405/commits/e1ab7d0b0850849a35d70acd2d3d0cdb95878abe, but do you also prefer to rename the `client` configuration directive under `listen` ?

src/main.c
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/a512ac174240ded47c5ed5546882a017ef22649e. 

>  I think we'd be fine using int instead (see max-connections).

This is something I've always wondered about. I understand that 64bit doesn't need to be specified, but why use int instead of unsigned? Are negative values expected?

src/main.c
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/a512ac174240ded47c5ed5546882a017ef22649e

lib/common/socketpool.c
i110: Done in https://github.com/h2o/h2o/pull/3405/commits/5be0daa4f070ad22b7313452ad92809882961257

src/main.c
i110: Yes as you mentioned socketpool can be shared among the threads, but my intention here was we does not share the sockpool for reverse tunnel, because we don't want to reuse sockets etc - then I noticed that we never reuse the sockets, and are just using socketpool as a connector, so we _can_ share it and there's no reason to have dedicated sockpool for each threads. Will fix soon

lib/common/reverse_tunnel.c
```diff
@@ -0,0 +1,189 @@
...
+{
+    h2o_reverse_ctx_t *reverse = (void *)sock->data;
+
+    h2o_socket_read_stop(sock);
+    if (err != NULL) {
+        h2o_error_printf("unexpected read error in on_reverse_rea: %s\n", err);
+        return;
+    }
+
+    h2o_accept(reverse->accept_ctx, sock);
```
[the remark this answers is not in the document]
i110: > we probably want to call h2o_http1_accept or h2o_http2_accept directly from here,

Thanks, agree with this

> depending on the value of the ALPN request header field

Is it `Selected-ALPN` response header field?

> also, we have to make the decision when we send the response headers, including the ALPN response header to indicate the protocol being chosen

I totally forgot about ALPN **request** header, and am willing to add it
NOTE: again, this is reverse-tunnel **client** implementation, not server.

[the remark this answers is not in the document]
i110: > I'm not sure if that is the correct thing to do with protocols like HTTP/2 that allows both endpoints start talking at the same time.
I think, in the case of http2, we may have to do something similar to `on_upgrade_complete` in lib/http2/connection.c, right?

> Maybe it is a sign that the approach is incorrect.

Instead of doing this in the socket read callback, can we do it in on_head and return a signal with semantics like `h2o_httpclient_socket_has_been_already_stolen_do_not_touch_it_anymore`?

lib/common/httpclient.c
kazuho: A function is a constant pointer (in such that you cannot alter the address or the contents).

src/main.c
kazuho: > do you also prefer to rename the client configuration directive under listen ?

That might make things less confusing, I think (we know the terms "client" and "server" is confusing due to us reversing the roles).

src/main.c
kazuho: The problem descibed in #2913 is as follows:
* Once we start accepting connections and receive SIGTERM, we need to shutdown gracefully. In order to do so, we broadcast the message using `h2o_multithread_send_message`.
* But we cannot call`h2o_multithread_send_message` before all the threads have completed setting up the receiver.
* Therefore, installation of the signal handler must happen after all the threads have completed their setup phase, but before they start accepting the connections.

lib/common/reverse_tunnel.c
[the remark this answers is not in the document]
kazuho: > I think, in the case of http2, we may have to do something similar to `on_upgrade_complete` in lib/http2/connection.c, right?

I assume you are pointing to the fact that in `on_upgrade_complete` we are consuming the bytes in the receive side and building HTTP/1.1 response headers of our own. Yes, that is one way forward.

> Instead of doing this in the socket read callback, can we do it in on_head and return a signal with semantics like `h2o_httpclient_socket_has_been_already_stolen_do_not_touch_it_anymore`?

First of all, I still think that it is correct to return a special value. I assume you think the same. Then, the question is if:
* we want to pass the size of the response header to the `on_head` callback so that it can steal the socket immediately, or
* we want to let the `on_head` set a callback (inside `h2o_httpclient_properties_t`) that will be invoked by httpclient once httpclient cosumes the response headers.

I lean towards the former.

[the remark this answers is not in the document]
i110: Another option would be: passing `h2o_socket_t *(*steal_socket)(void)` function pointer as a field of h2o_httpclient_properties, and the consumer calls it if it want to steal the socket. By doing so we can write buffer consuming code in it, and as an additional bonus, we no longer need `h2o_httpclient_steal_socket` return value to signal. Of course it can be a public function like `h2o_socket_t *h2o_httpclient_steal_socket(void)` but it might be a bit complex to dispatch to the actual client implementations (h1/h2/h3).

src/main.c
i110: Done in https://github.com/h2o/h2o/pull/3405/commits/a10d7bc3407f99cf4e10d50cf3df769d46dc56d1

lib/common/httpclient.c
i110: sorry what do you mean?

src/main.c
i110: While I was making the changes, I started to lose track of the benefits of these changes. The caller of `h2o_reverse_init` is not interested in what modules the reverse tunnel is using internally ex: httpclient, and socketpool. It seems simpler and easier for me to understand to just pass the client url and SSL_CTX. I agree it feels a bit odd not to share what can be shared, but in reality, the socketpool only serves as a connector in this case, and creating it individually doesn't seem to cause any real issues. Could you give me a more convincing explanation of why it's better to share the socketpool instead of ssl_ctx?

lib/common/reverse_tunnel.c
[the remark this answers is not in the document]
kazuho: Yeah, I think I prefer returning a special value, as that would make things more explicit in the source code of http1client.c. We _need_ to return something, therefore the cost of having a signal is marginal.

What I am concerned is that pointers to the response being retained by http1client becomes invalid the moment the socket receive buffer is consumed. If that is going to happen at a different moment for stolen socket, that should be easy to understand when reading http1client.c. Otherwise, we might break things.

src/main.c
i110: Thank you for the explanation. And reading your comment I remember that I have to add graceful shutdown handling also for reverse-listening connections.

i110: Oh, it looks like I had already marked that as TODO :yum:
https://github.com/h2o/h2o/pull/3405/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176R4353

lib/common/reverse_tunnel.c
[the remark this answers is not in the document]
i110: ok then let's have `rlen` in `h2o_httpclient_conn_properties_t`.

By the way, just to confirm, another obvious solution is setting zero-timeout (deferred) callback in `on_head` and do all the things in that callback, but you don't like that, do you?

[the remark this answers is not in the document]
kazuho: Having rlen (or `size_t *rlen` so that the application cannot steal H2 and H3 connectsions) sounds good to me.

> By the way, just to confirm, another obvious solution is setting zero-timeout (deferred) callback in `on_head` and do all the things in that callback, but you don't like that, do you?

Yeah I think doing that just makes maintenance harder. I think what we want is a clear cut and explicit transition of ownership of the socket (from http1client to the app). The idea of using a deferred callback to give http1client the time to consumes the bytes is exactly the opposite.

src/main.c
kazuho: Socket pools are not only about pooling the sockets. It's a generic structure describing how connections should be made by httpclient.

Obviously, `SSL_CTX` is the most prominent one. But there are also things like load balancer and target list that we might want to exploit in the future.

Or taking a step back, the question would be why we need to make the socket pool internal to reverse connect, when the reverse proxy exposes that for configuration? I do not see a reason to introduce inconsistency here.

i110: Thank you for the explanation. I made it pass socketpool instead of ssl_ctx in https://github.com/h2o/h2o/pull/3405/commits/f113045ff6488995a99bbc97f54937ea6ea8b7e7.

lib/common/reverse_tunnel.c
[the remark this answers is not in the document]
i110: Just added `steal_bytes` to conn props and set rlen value in https://github.com/h2o/h2o/pull/3405/commits/216b991f6bbabed67653ef28dccc03e3f953e5b6

src/main.c
```diff
@@ -4215,16 +4393,18 @@ H2O_NORETURN static void *run_loop(void *_thread_index)
...
+        if (!is_quic_listener(conf.listeners[i]) && !is_reverse_listener(conf.listeners[i]))
             h2o_socket_read_stop(listeners[i].sock);
     }
     h2o_evloop_run(conf.threads[thread_index].ctx.loop, 0);
     for (i = 0; i != conf.num_listeners; ++i) {
-        if (conf.listeners[i]->quic.ctx == NULL) {
+        if (is_quic_listener(conf.listeners[i])) {
+            listeners[i].http3.ctx.super.acceptor = NULL;
+        } else if (is_reverse_listener(conf.listeners[i])) {
+            // TODO
```
i110, line 4404: perhaps we don't need to do anything

include/h2o.h
```diff
@@ -2356,6 +2380,16 @@ void h2o_headers_append_command(h2o_headers_command_t **cmds, int cmd, h2o_heade
...
 
+
+/**
+ * extract header name from given string
+ */
+int h2o_headers_extract_name(const char *src, size_t len, h2o_iovec_t **name);
+/**
+ * extract header name and value from given string
+ */
+int h2o_headers_extract_name_value(const char *src, size_t len, h2o_iovec_t **name, h2o_iovec_t *value);
```
kazuho, line 2391: I think it makes sense to reuse these functions.

But the names are confusing, as they are not dealing with headers (list of headers). How about moving them to string_.h / string.c, and calling them `h2o_extract_header_name`, `h2o_extract_header_name_and_value`.

Also, I would appreciate writing down what the return values mean and what name and values represent (are they allocated? are they tokens?).

lib/common/http1client.c
```diff
@@ -553,13 +556,21 @@ static void on_head(h2o_socket_t *sock, const char *err)
         on_head.pipe_reader = &client->pipe_reader;
 #endif
 
+    client->steal_bytes = (size_t)rlen;
```
kazuho, line 559: Rather than adding `steal_bytes` as a member of `h2o_http1client_t`, how about passing `rlen` as an argument to `call_on_head`? By doing so, the code becomes clearer regarding when the value can be non-zero.

lib/common/httpclient.c
kazuho: Regarding your argument that it is good to be explicit that something is a constant pointer, my point is that a function declaration is exactly about declaring a constant pointer.

src/main.c
kazuho: I think it comes from the fact that `sync_fetch_and_add` is a pointer and a value of same type, and that we want to pass `-1` as the second argument (see how `num_connections(-1)` is used).

If we change `num_connections` to an unsigned type, we would have to worry how `-1` is promoted to an unsigned type (and that expansion does not happen after the sign conversion). That's a needless complexity IMO.

kazuho (commented):
Thank you for the changes. I think we are making great progress.

src/main.c
```diff
@@ -2632,28 +2701,31 @@ static void on_http3_conn_destroy(h2o_quic_conn_t *conn)
...
+        yoml_t **port_node, **host_node, **type_node, **proxy_protocol_node, **sndbuf_node, **rcvbuf_node, **reconnect_interval_node, **connections_per_thread_node;
         if (h2o_configurator_parse_mapping(
-                cmd, node, "port:s",
-                "host:s,type:s,owner:s,group:s,permission:*,ssl:m,proxy-protocol:*,quic:m,cc:s,initcwnd:s,sndbuf:s,rcvbuf:s",
-                &port_node, &host_node, &type_node, &owner_node, &group_node, &permission_node, &ssl_node, &proxy_protocol_node,
-                &quic_node, &cc_node, &initcwnd_node, &sndbuf_node, &rcvbuf_node) != 0)
+                cmd, node, NULL,
+                "host:s,port:s,type:s,owner:s,group:s,permission:*,ssl:m,proxy-protocol:*,quic:m,cc:s,initcwnd:s,sndbuf:s,rcvbuf:s,url:s,reconnect-interval:s,connections-per-thread:s,header:*",
+                &host_node, &port_node, &type_node, &owner_node, &group_node, &permission_node, &ssl_node, &proxy_protocol_node,
+                &quic_node, &cc_node, &initcwnd_node, &sndbuf_node, &rcvbuf_node, &url_node, &reconnect_interval_node, &connections_per_thread_node, &header_node) != 0)
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/3ea2261f9f734847d3968ac43844e0d23e0e8c87

src/main.c
```diff
@@ -222,6 +222,15 @@ struct listener_config_t {
          */
         struct listener_config_t *sibling;
     } quic;
+
+    struct {
+        h2o_url_t url;
+        uint64_t reconnect_interval;
+        uint64_t connections_per_thread;
+        h2o_headers_t req_headers;
```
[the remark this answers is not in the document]
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/2c313750cc26715389a0bee41bc86c3c19d27364

lib/core/headers.c
```diff
@@ -41,7 +41,12 @@ static ssize_t add_header(h2o_mem_pool_t *pool, h2o_headers_t *headers, h2o_iove
 
 static inline h2o_iovec_t *alloc_and_init_iovec(h2o_mem_pool_t *pool, const char *base, size_t len)
 {
-    h2o_iovec_t *iov = h2o_mem_alloc_pool(pool, *iov, 1);
+    h2o_iovec_t *iov;
+    if (pool != NULL) {
+        iov = h2o_mem_alloc_pool(pool, *iov, 1);
+    } else {
+        iov = h2o_mem_alloc(sizeof(*iov));
+    }
```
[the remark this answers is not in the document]
i110: Also addressed in https://github.com/h2o/h2o/pull/3405/commits/2c313750cc26715389a0bee41bc86c3c19d27364

include/h2o.h
i110: Addressed in https://github.com/h2o/h2o/pull/3405/commits/5bb1798f9539722d43f30e582bc48df86a46d31a

lib/common/http1client.c
i110: > how about passing rlen as an argument to call_on_head? By doing so, the code becomes clearer regarding when the value can be non-zero.

Addressed in https://github.com/h2o/h2o/pull/3405/commits/494436bf0dc27f3743fb7d48a1873ed44fbfaf55, but

> Rather than adding steal_bytes as a member of h2o_http1client_t

We still need to store that rlen value as a member of client, because we should provide it in get_conn_properties callback.

PR #3409, opened 2024-09-04
When h2c is disabled, redirect to https

deweerdt:
The current behavior lets the request through and depending on whether the handler lets an `upgrade:` request through or not, might forward the request and attempt an upgrade or return a 403.

This PR makes h2o redirect to https instead.


deweerdt:
discussed this offline, we'll want to let the request through. Closing to re-open.

PR #3421, opened 2024-10-15
Reject a valid but problematic configuration when H2O is starting: handler reordering due to deferral, request streaming mismatch.

jhatala:
Two checks that reject a valid but problematic configuration when H2O is starting rather than leading to silent problems later:

1\. Reject a configuration with a misleading order of handlers or filters.  With this change it becomes an error if a configuration command that adds a deferred handler appears in the configuration file before a command that adds a semi-deferred or non-deferred handler, or a semi-deferred one before a non-deferred one.

Some configurations that were valid before this change are now rejected.  In a case like that, the admin will need to re-order the directives so that they are no longer in misleading order.

2\. Optionally, if `request-body-streaming: ON` appears in a path-conf, reject a configuration with a handler for that path-conf that does not set the `supports_request_streaming` flag.

This does not affect existing configuration because it requires an opt-in marker in the form of `request-body-streaming: ON` anywhere in the path-conf.

The function `h2o_req_can_stream_request` looks at the first handler applicable to the request to decide whether to buffer the request body or to invoke the handler without buffering the request body.  If the admin configures H2O with a non-streaming handler like `mruby.handler` in front of a streaming like `proxy.reverse.url` , they inadvertently lose the request body streaming property.  In the reverse case, if the first handler is a streaming handler and it declines to a non-streaming handler, then the non-streaming handler will run with a missing or incomplete `req->entity`.

This change simply detects a potential problem like this when H2O starts.  For the streaming handler mismatch case, the check requires an opt-in in the configuration, in order for the admin to express their intent to use the request body streaming functionality of their streaming-capable handler.

I think that it would improve configuration safety if `h2o_req_can_stream_request` only returned true if all applicable handlers support streaming, not just the first one, but that is not part of this PR.


jhatala:
`t/60find-doc.t` flagged me

jhatala:
documentation added in b15e9e5f0 and regen 70edfa106

jhatala:
> I think that it would improve configuration safety if h2o_req_can_stream_request only returned true if all applicable handlers support streaming, not just the first one, but that is not part of this PR.

=> #3423

PR #3455, opened 2025-01-10
If the socketpool's timeout is 0, do not lookup a socket

deweerdt:
This avoids lock contention in cases where the idle timeout is 0, and sockets aren't going to be re-used anyway. It is restricted to non-global pools (i.e. the ones created by `proxy.reverse.url`)  because the global pool can be resized concurrently.


lib/common/socketpool.c
```diff
@@ -471,6 +471,9 @@ void h2o_socketpool_connect(h2o_socketpool_connect_request_t **_req, h2o_socketp
     size_t target = SIZE_MAX;
     h2o_linklist_t *sockets = NULL;
 
+    if (pool->timeout == 0 && !h2o_socketpool_is_global(pool))
```
kazuho, line 474: IIUC, the rationale behind skipping is that when `pool->timeout` is zero, no socket would have been added to the pool. Assuming that is the case:
* Do we need to call `h2o_socketpool_is_global` here? It is just a type of pool.
* Do we need to make changes to `h2o_socketpool_return` too? I think that function takes a lock, adds the socket to the pool, then destroys the ones that have expired.

kazuho (commented):
Thank you for the PR. The change by itself looks alright but I have questions:

lib/common/socketpool.c
kazuho, line 474:
```suggestion
    if (pool->timeout == 0)
```

kazuho: I've addressed the first point is addressed in 03bf672.

Regarding the second point, I notice that while it is true, it is unneeded by the proxy handler, because if `pool->timeout == 0` then `h2o_socketpool_can_keepalive` return false, indicating that the HTTP connection established to the backends will avoid doing keep alives.

kazuho:
After all, the code as written by @deweerdt was 100% correct and the justifications were available in the original issue above.

Sorry for the fuss. I've added code comments that justifies the approach.

PR #3462, opened 2025-01-30
add support for busypolling

sleepybishop:
This pr add busypolling support to h2o. It is linux only.

For notes on busy polling and it's internals, please see the following:

  - https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/tree/tools/testing/selftests/net/busy_poller.c
  - https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/tree/Documentation/networking/napi.rst

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,27 @@ ELSE ()
 ENDIF ()
 OPTION(WITH_KTLS "use Kernel TLS" ${WITH_KTLS_DEFAULT})
 
+# Find out if system provides libynl
+CMAKE_PUSH_CHECK_STATE()
+FIND_LIBRARY(SYSTEM_YNL_LIB "ynl")
+IF (SYSTEM_YNL_LIB)
+    SET(CMAKE_REQUIRED_LIBRARIES ${SYSTEM_YNL_LIB})
+ENDIF()
+CHECK_C_SOURCE_COMPILES("
+#include <ynl/ynl.h>
+int main(void) {
+return YNL_ERROR_NONE;
+}" SYSTEM_HAS_YNL)
+CMAKE_POP_CHECK_STATE()
+
+IF (SYSTEM_HAS_YNL)
+    ADD_DEFINITIONS("-DH2O_HAS_YNL_H")
+    IF (SYSTEM_YNL_LIB)
+        LIST(APPEND EXTRA_LIBS ${SYSTEM_YNL_LIB})
+    ENDIF ()
+ENDIF ()
+
+
 # aegis
 FIND_PACKAGE(aegis)
 IF (aegis_FOUND)
@@ -405,6 +426,7 @@ SET(LIB_SOURCE_FILES
     lib/common/balancer/roundrobin.c
     lib/common/balancer/least_conn.c
     lib/common/absprio.c
+    lib/common/busypoll.c
 
     lib/core/config.c
     lib/core/configurator.c
@@ -441,6 +463,7 @@ SET(LIB_SOURCE_FILES
     lib/handler/status/ssl.c
     lib/handler/status/durations.c
     lib/handler/configurator/access_log.c
+    lib/handler/configurator/busypoll.c
     lib/handler/configurator/compress.c
     lib/handler/configurator/errordoc.c
     lib/handler/configurator/expires.c
--- a/include/h2o.h
+++ b/include/h2o.h
@@ -52,6 +52,7 @@ extern "C" {
 #include "h2o/token.h"
 #include "h2o/url.h"
 #include "h2o/balancer.h"
+#include "h2o/busypoll.h"
 #include "h2o/http2_common.h"
 #include "h2o/send_state.h"
 
@@ -583,6 +584,12 @@ struct st_h2o_globalconf_t {
     h2o_status_callbacks_t statuses;
 
     size_t _num_config_slots;
+
+    /* busypoll */
+    struct {
+        size_t nic_count;
+        h2o_busypoll_nic_vector_t nic_to_cpu_map;
+    } bp;
 };
 
 enum {
@@ -663,6 +670,11 @@ typedef enum h2o_conn_state {
  * context of the http server.
  */
 struct st_h2o_context_t {
+    /**
+     * thread index/count, used in status handlers for per-thread metrics
+     */
+    int thread_index;
+    int thread_count;
     /**
      * points to the loop (either uv_loop_t or h2o_evloop_t, depending on the value of H2O_USE_LIBUV)
      */
@@ -2025,6 +2037,9 @@ void h2o_access_log_register_configurator(h2o_globalconf_t *conf);
 /* lib/handler/server_timing.c */
 void h2o_server_timing_register(h2o_pathconf_t *pathconf, int enforce);
 void h2o_server_timing_register_configurator(h2o_globalconf_t *conf);
+
+/* lib/common/busypoll.c */
+void h2o_busypoll_register_configurator(h2o_globalconf_t *conf);
 
 /* lib/compress.c */
 
--- a/include/h2o/busypoll.h
+++ b/include/h2o/busypoll.h
@@ -1,0 +1,34 @@
+#ifndef h2o__busypoll_h
+#define h2o__busypoll_h
+
+#include <stdint.h>
+
+enum busypoll_mode_t {
+    BP_MODE_OFF = 0,
+    BP_MODE_SUSPEND,
+    BP_MODE_BUSYPOLL,
+};
+
+struct busypoll_nic_t {
+    h2o_iovec_t iface;
+    size_t ifindex;
+    size_t cpu_count;
+    enum busypoll_mode_t mode;
+    cpu_set_t cpu_map;
+    pthread_mutex_t mutex;
+    struct {
+        size_t gro_flush_timeout;
+        size_t defer_hard_irqs;
+        size_t suspend_timeout;
+    } options;
+};
+
+typedef H2O_VECTOR(struct busypoll_nic_t) h2o_busypoll_nic_vector_t;
+
+void h2o_busypoll_bind_interface(int fd, const char *iface);
+void h2o_busypoll_attach_cbpf(int fd, uint16_t cpus, const char *iface);
+void h2o_busypoll_set_opts(uint32_t ifindex, uint32_t defer_hard_irqs, uint64_t gro_flush_timeout, uint64_t irq_suspend_timeout);
+void h2o_busypoll_handle_nic_map_accept(h2o_socket_t *sock, h2o_socket_t *listener, size_t thread_index,
+                                        struct busypoll_nic_t *nic_to_cpu_map, size_t nic_count);
+
+#endif
--- a/include/h2o/socket/evloop.h
+++ b/include/h2o/socket/evloop.h
@@ -22,6 +22,7 @@
 #ifndef h2o__evloop_h
 #define h2o__evloop_h
 
+#include <stdbool.h>
 #include "h2o/linklist.h"
 #include "h2o/timerwheel.h"
 
@@ -53,6 +54,12 @@ typedef struct st_h2o_evloop_t {
     h2o_timerwheel_t *_timeouts;
     h2o_sliding_counter_t exec_time_nanosec_counter;
     uint64_t run_count;
+    struct {
+        uint64_t epoll_bp_usecs;
+        uint64_t epoll_bp_budget;
+        bool epoll_bp_changed;
+        bool epoll_nonblock;
+    } bp;
 } h2o_evloop_t;
 
 typedef h2o_evloop_t h2o_loop_t;
@@ -71,6 +78,7 @@ h2o_socket_t *h2o_evloop_socket_accept(h2o_socket_t *listener);
 void h2o_evloop_socket_set_max_read_size(h2o_socket_t *sock, size_t max_size);
 
 h2o_evloop_t *h2o_evloop_create(void);
+h2o_evloop_t *h2o_evloop_create_busy_poll(uint64_t nsecs, uint64_t budget);
 void h2o_evloop_destroy(h2o_evloop_t *loop);
 /**
  * runs a event loop once. The function returns 0 if successful, or -1 if it aborts the operation due to a system call returning an
@@ -99,6 +107,27 @@ static inline uint64_t h2o_now(h2o_evloop_t *loop)
 static inline uint64_t h2o_now_nanosec(h2o_evloop_t *loop)
 {
     return loop->_now_nanosec;
+}
+
+static inline void h2o_loop_set_bp_usecs(h2o_evloop_t *loop, uint64_t usecs)
+{
+    if (loop->bp.epoll_bp_usecs != usecs) {
+        loop->bp.epoll_bp_usecs = usecs;
+        loop->bp.epoll_bp_changed = true;
+    }
+}
+
+static inline void h2o_loop_set_bp_budget(h2o_evloop_t *loop, uint64_t budget)
+{
+    if (loop->bp.epoll_bp_budget != budget) {
+        loop->bp.epoll_bp_budget = budget;
+        loop->bp.epoll_bp_changed = true;
+    }
+}
+
+static inline void h2o_loop_set_nonblock(h2o_evloop_t *loop, bool nonblock)
+{
+    loop->bp.epoll_nonblock = nonblock;
 }
 
 static inline uint64_t h2o_evloop_get_execution_time_millisec(h2o_evloop_t *loop)
--- a/lib/common/busypoll.c
+++ b/lib/common/busypoll.c
@@ -1,0 +1,444 @@
+#include "h2o.h"
+#include "h2o/busypoll.h"
+
+#include <errno.h>
+#include <error.h>
+
+#if defined(__linux__) && defined(SO_REUSEPORT) && defined(H2O_HAS_YNL_H)
+#include <linux/genetlink.h>
+#include <linux/netlink.h>
+#include <linux/filter.h>
+#include <linux/if.h>
+#include <sys/ioctl.h>
+#include <sys/sysinfo.h>
+#include <ynl/ynl.h>
+#include <ynl/netdev-user.h>
+
+static __thread int cpu_claimed = 0;
+static __thread const char *nic_claimed = NULL;
+static __thread unsigned int napi_id_claimed = 0;
+
+static void setup_queue(uint32_t cfg_ifindex, uint32_t cfg_defer_hard_irqs, uint64_t cfg_gro_flush_timeout,
+                        uint64_t cfg_irq_suspend_timeout)
+{
+    struct netdev_napi_get_list *napi_list = NULL;
+    struct netdev_napi_get_req_dump *req = NULL;
+    struct netdev_napi_set_req *set_req = NULL;
+    struct ynl_sock *ys;
+    struct ynl_error yerr;
+    uint32_t napi_id = 0;
+
+    ys = ynl_sock_create(&ynl_netdev_family, &yerr);
+    if (!ys)
+        error(1, 0, "YNL: %s", yerr.msg);
+
+    req = netdev_napi_get_req_dump_alloc();
+    netdev_napi_get_req_dump_set_ifindex(req, cfg_ifindex);
+    napi_list = netdev_napi_get_dump(ys, req);
+
+    /* assume there is 1 NAPI configured and take the first */
+    if (napi_list->obj._present.id)
+        napi_id = napi_list->obj.id;
+    else
+        error(1, 0, "napi ID not present?");
+
+    set_req = netdev_napi_set_req_alloc();
+    netdev_napi_set_req_set_id(set_req, napi_id);
+    netdev_napi_set_req_set_defer_hard_irqs(set_req, cfg_defer_hard_irqs);
+    netdev_napi_set_req_set_gro_flush_timeout(set_req, cfg_gro_flush_timeout);
+    netdev_napi_set_req_set_irq_suspend_timeout(set_req, cfg_irq_suspend_timeout);
+
+    if (netdev_napi_set(ys, set_req))
+        error(1, 0, "can't set NAPI params: %s\n", yerr.msg);
+
+    netdev_napi_get_list_free(napi_list);
+    netdev_napi_get_req_dump_free(req);
+    netdev_napi_set_req_free(set_req);
+    ynl_sock_destroy(ys);
+}
+
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
+#endif
+
+void attach_cbpf(int fd, uint16_t cpus, const char *iface)
+{
+    struct sockaddr_storage ss;
+    socklen_t sslen = sizeof(ss);
+    uint16_t n = cpus;
+    int ret = 0;
+
+    ret = getsockname(fd, (struct sockaddr *)&ss, &sslen);
+    if (ret != 0) {
+        fprintf(stderr, "getsockname failed on listener: %s\n", strerror(errno));
+    }
+    if (ss.ss_family == AF_UNIX) {
+        fprintf(stderr, "dont add reuseport filter to AF_UNIX socket listener\n");
+        return;
+    } else if (ss.ss_family == AF_INET) {
+        fprintf(stderr, "will add filter to AF_INET socket\n");
+    } else if (ss.ss_family == AF_INET6) {
+        fprintf(stderr, "will add filter to AF_INET6 socket\n");
+    } else {
+        fprintf(stderr, "unknown family: %lu, adding\n", (unsigned long)ss.ss_family);
+    }
+
+    struct sock_filter code[] = {
+        /* A = skb->queue_mapping */
+        {BPF_LD | BPF_W | BPF_ABS, 0, 0, SKF_AD_OFF + SKF_AD_QUEUE},
+        /* A = A % n */
+        {BPF_ALU | BPF_MOD, 0, 0, n},
+        /* return A */
+        {BPF_RET | BPF_A, 0, 0, 0},
+    };
+
+    struct sock_fprog p = {
+        .len = ARRAY_SIZE(code),
+        .filter = code,
+    };
+
+    if (setsockopt(fd, SOL_SOCKET, SO_ATTACH_REUSEPORT_CBPF, &p, sizeof(p))) {
+        char buf[128];
+        h2o_fatal("failed to set SO_ATTACH_REUSEPORT_CBPF: %d:%s", errno, h2o_strerror_r(errno, buf, sizeof(buf)));
+    } else {
+        fprintf(stderr, "bpf prog attached to fd:%d, queues: %u\n", fd, cpus);
+    }
+
+    return;
+}
+
+void bind_interface(int fd, const char *iface)
+{
+    /* setting SO_BINDTODEVICE causes listen sockets with the same
+     * attributes (address, port, nic index) to be part of the same
+     * reuseport group.
+     *
+     * this is critical because if the nic index is not set with
+     * BINDTODEVICE, the reuseport group will receive incoming connections
+     * from any NIC and the cBPF program which does a modulo on the
+     * queue_mapping will fail (since multiple NICs each can have the same
+     * queue mapping)
+     */
+    if (iface && setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, iface, strlen(iface)) == -1) {
+        h2o_perror("failed to SO_BINDTODEVICE");
+        abort();
+    }
+
+    if (iface) {
+        fprintf(stderr, "bound %d to iface: %s\n", fd, iface);
+    }
+}
+
+static unsigned int get_napi_id(int fd)
+{
+#if !defined(SIOCGIFNAME_BY_NAPI_ID)
+#define SIOCGIFNAME_BY_NAPI_ID (0x894D)
+#endif
+    unsigned int napi_id = 0;
+    socklen_t napi_id_len = sizeof(napi_id);
+
+    if (getsockopt(fd, SOL_SOCKET, SO_INCOMING_NAPI_ID, &napi_id, &napi_id_len) != 0) {
+        h2o_perror("so_napi_incoming_id failed");
+    }
+
+    return napi_id;
+}
+
+static int get_nic_name_by_napi(int fd, unsigned int napi_id, struct ifreq *ifr)
+{
+    ifr->ifr_ifru.ifru_ivalue = napi_id;
+
+    return ioctl(fd, SIOCGIFNAME_BY_NAPI_ID, ifr);
+}
+
+/*
+ * -2  error
+ * -1  unix socket
+ *  0  non-local INET (or INET6) connection
+ *  1  local INET (or INET6) connection
+ */
+static int is_local_conn(int fd)
+{
+    int is_local = -2;
+    int domain = AF_UNSPEC;
+    socklen_t domain_size = sizeof(domain);
+    int r = getsockopt(fd, SOL_SOCKET, SO_DOMAIN, &domain, &domain_size);
+    if (r < 0) {
+        h2o_perror("getsockopt SO_DOMAIN failed");
+        return -2;
+    }
+    if (domain_size != sizeof(domain)) {
+        h2o_perror("getsockopt SO_DOMAIN length mismatch");
+        return -2;
+    }
+
+    /* unix sockets are not the sockets we are looking for */
+    if (domain == AF_UNIX || domain == AF_LOCAL) {
+        return -1;
+    }
+
+    return is_local;
+}
+
+static int assign_nic_map_cpu(const char *iface, size_t thread_index, int napi_id, int listener_fd, int is_lo,
+                              struct busypoll_nic_t *nic_to_cpu_map, size_t nic_count)
+{
+    int found = 0;
+    for (int i = 0; i < nic_count; i++) {
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+
+        if (strcmp(nic_to_cpu_map[i].iface.base, iface) == 0) {
+            fprintf(stderr, "found matching iface in conf map: %s\n", iface);
+            pthread_mutex_lock(&nic_to_cpu_map[i].mutex);
+            for (int j = 0; j < get_nprocs_conf(); j++) {
+
+                if (CPU_ISSET(j, &nic_to_cpu_map[i].cpu_map)) {
+                    /* take the first CPU that is set, and clear it so other threads can't have it */
+                    CPU_CLR(j, &nic_to_cpu_map[i].cpu_map);
+
+                    /* set the CPU in this threads temporary set so the thread can pin itself */
+                    CPU_SET(j, &cpuset);
+                    pthread_setaffinity_np(pthread_self(), sizeof(cpuset), &cpuset);
+
+                    /* if using neverbleed, the neverbleed threads inherit
+                     * their CPU set from the default h2o mask (which
+                     * excludes busy poll CPUs).
+                     *
+                     * all that is left to do is ensure that yield (sleep)
+                     * on read is disabled.
+                     */
+
+                    /* FIXME:
+                    if (neverbleed != NULL) {
+                        set_yield_on_read(neverbleed, 0);
+                    }
+                    */
+
+                    fprintf(stderr, "thread idx %zd, has taken cpu: %d, for NIC: %s, via napi: %d via listener %d\n", thread_index,
+                            j, nic_to_cpu_map[i].iface.base, napi_id, listener_fd);
+
+                    /* we can print which listener the accept arrived on to
+                     * help debug / figure out why this thread took this
+                     * CPU.
+                     *
+                     * some ifaces, like lo, pass listener_fd = 0, so there
+                     * would be nothing to print.
+                     */
+                    if (listener_fd > 0) {
+                        int ret;
+                        struct sockaddr_storage ss;
+                        socklen_t len = sizeof(ss);
+                        ret = getsockname(listener_fd, (struct sockaddr *)&ss, &len);
+                        if (ret != 0) {
+                            fprintf(stderr, "getsockname failed on listener: %s\n", strerror(errno));
+                        }
+                    }
+
+                    nic_claimed = nic_to_cpu_map[i].iface.base;
+                    napi_id_claimed = napi_id;
+                    found = 1;
+                    cpu_claimed = 1;
+                    break;
+                }
+            }
+            pthread_mutex_unlock(&nic_to_cpu_map[i].mutex);
+            if (found)
+                break;
+        }
+    }
+
+    return found;
+}
+
+static void handle_nic_map_accept(h2o_socket_t *sock, h2o_socket_t *listener, size_t thread_index,
+                                  struct busypoll_nic_t *nic_to_cpu_map, size_t nic_count)
+{
+    /* this would probably be set per-listener or something as sndbuf and rcvbuf are below ? */
+    int sockfd = h2o_socket_get_fd(sock);
+    unsigned int napi_id = 0;
+    int prio = 1;
+    if (setsockopt(sockfd, SOL_SOCKET, SO_PRIORITY, &prio, sizeof(prio)) != 0) {
+        h2o_perror("so_priority failed");
+    }
+
+    napi_id = get_napi_id(sockfd);
+
+    /* if this thread has claimed a CPU, check the incoming connection to make sure it is correct */
+    if (cpu_claimed) {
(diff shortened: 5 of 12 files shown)


lib/common/busypoll.c
```diff
@@ -0,0 +1,444 @@
...
+    socklen_t napi_id_len = sizeof(napi_id);
+
+    if (getsockopt(fd, SOL_SOCKET, SO_INCOMING_NAPI_ID, &napi_id, &napi_id_len) != 0) {
+        h2o_perror("so_napi_incoming_id failed");
+    }
+
+    return napi_id;
+}
+
+static int get_nic_name_by_napi(int fd, unsigned int napi_id, struct ifreq *ifr)
```
jdamato-fsly, line 147: this function (and its callers) should be re-written -- this ioctl is just a hack, not a thing that exists in the real public linux kernel.

instead, when h2o starts it should compute the mapping of interfaces and NAPI IDs, and the implementation of this function would take a NAPI ID and return the interface name from the map h2o has computed

lib/common/busypoll.c
```diff
@@ -0,0 +1,444 @@
...
+#include <sys/ioctl.h>
+#include <sys/sysinfo.h>
+#include <ynl/ynl.h>
+#include <ynl/netdev-user.h>
+
+static __thread int cpu_claimed = 0;
+static __thread const char *nic_claimed = NULL;
+static __thread unsigned int napi_id_claimed = 0;
+
+static void setup_queue(uint32_t cfg_ifindex, uint32_t cfg_defer_hard_irqs, uint64_t cfg_gro_flush_timeout,
```
jdamato-fsly, line 21: i think this probably needs to take a NAPI ID, as these options should be set on a per-NAPI basis... the implementation below seems to set them only on the first queue

lib/common/busypoll.c
```diff
@@ -0,0 +1,444 @@
...
+}
+
+void h2o_busypoll_bind_interface(int fd, const char *iface)
+{
+    bind_interface(fd, iface);
+}
+
+void h2o_busypoll_set_opts(uint32_t ifindex, uint32_t defer_hard_irqs, uint64_t gro_flush_timeout, uint64_t irq_suspend_timeout)
+{
+    setup_queue(ifindex, defer_hard_irqs, gro_flush_timeout, irq_suspend_timeout);
```
jdamato-fsly, line 406: this should probably either iterate across the NAPI IDs for an interface setting the options, or `setup_queue` internally should iterate across a range of queue numbers setting these settings

src/main.c
```diff
@@ -2605,6 +2601,10 @@ static int finish_open_listener(int fd, int apply_bpf_filter, int cpus, const ch
     if (fd < 0)
         goto Error;
 
+    if (conf.globalconf.bp.nic_count > 0) {
+        h2o_busypoll_bind_interface(fd, iface);
```
jdamato-fsly, line 2605: this change on its own is likely not going to be sufficient -- you'll need to rework the logic in `open_listener` to move more of it to after this call in `finish _open_listener`

lib/handler/configurator/busypoll.c
```diff
@@ -0,0 +1,130 @@
...
+        switch (h2o_configurator_get_one_of(cmd, *mode_node, "OFF,SUSPEND,BUSYPOLL")) {
+        case 0:
+            nic_to_cpu_map->entries[i].mode = BP_MODE_OFF;
+            break;
+        case 1:
+            nic_to_cpu_map->entries[i].mode = BP_MODE_SUSPEND;
+            break;
+        case 2:
+#ifndef H2O_HAS_YNL_H
+            h2o_configurator_errprintf(cmd, node, "libynl is not available and required to busypoll");
```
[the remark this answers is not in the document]
sleepybishop: thank you, addressed in https://github.com/h2o/h2o/pull/3462/commits/bf340c7aed6c62875616e6c570b2bd614053c86b

src/main.c
sleepybishop: I've reverted this change in a53af73e0dc92341ce8af3d170b198851913350b, i don't think i can easily move the `bind()` out of `open_listener`, which means i can't move the setting of `so_reuseport` and since `so_bindtodevice` has to be set before that, it has to remain in `open_listener`

lib/common/busypoll.c
sleepybishop: thank you, i opted for the latter approach in 29542d8d3383aba2a4c59334be6de0f3c871202b. `setup_queue` iterates internally and saves the napi ids for later use.

lib/common/busypoll.c
sleepybishop: agree, see https://github.com/h2o/h2o/pull/3462#discussion_r1945900952

lib/common/busypoll.c
sleepybishop: thank you, i rewrote `get_nic_name_by_napi`. im at loss at how to reimplement `get_napi_id`, but im thinking we don't need to if we save the napi_id into the loop at assign time. WDYT?

lib/common/socket/evloop/epoll.c.h
```diff
@@ -196,9 +237,34 @@ int evloop_do_proceed(h2o_evloop_t *_loop, int32_t max_wait)
...
+            nsec = (_time % 1000000) * 1000;
+        }
+
+        struct timespec ts = {.tv_sec = seconds, .tv_nsec = nsec};
+        nevents = epoll_pwait2(loop->ep, events, sizeof(events) / sizeof(events[0]), &ts, NULL);
+    } else {
+        nevents = epoll_wait(loop->ep, events, sizeof(events) / sizeof(events[0]), max_wait);
+    }
+#else
     max_wait = adjust_max_wait(&loop->super, max_wait);
```
jdamato-fsly, line 264: i think `max_wait` needs to be moved up because it is used in the else case on line 261 above, but not computed.

include/h2o/socket/evloop.h
```diff
@@ -53,6 +54,13 @@ typedef struct st_h2o_evloop_t {
     h2o_timerwheel_t *_timeouts;
     h2o_sliding_counter_t exec_time_nanosec_counter;
     uint64_t run_count;
+    struct {
+        uint64_t epoll_bp_usecs;
+        uint64_t epoll_bp_budget;
+        bool prefer_busy_poll;
```
jdamato-fsly, line 60: i wasn't sure where to put this comment, but i think that `prefer_busy_poll` should also have a configuration option like the other two options.

the use case for this would be:
  - regular busy polling (e.g. `mode: BUSYPOLL`)
  - the user wants the `gro-flush-timeout` to be deferrred (which is only possible if prefer_busy_poll is enabled)

jdamato-fsly: or to say it a bit differently:

- `prefer_busy_poll` can optionally be used with `mode: BUSYPOLL`, but is not required.
- `prefer_busy_poll` must be set to 1 for `mode: SUSPEND` to be enabled

lib/common/socket/evloop/epoll.c.h
jdamato-fsly: in the case of `suspend`, i think there should be extra logic added to set `max_wait = -1` -- in `suspend` mode, the idea is that the polling happens in the kernel until an RX queue finds nothing, at which point IRQs are re-enabled and the process can be put to sleep (because there is no data to process). and it will be re-awoken when data arrives and an IRQ is generated.

so, in psuedocode maybe something like:

```
if (mode == suspend)
  max_wait = -1
else
  max_wait = adjust_max_wait(&loop->super, max_wait);
```

or maybe move that logic into `adjust_max_wait` itself and have it return `-1` if suspend mode is active for this `loop` ?

include/h2o/socket/evloop.h
sleepybishop: thank you, addressed in eae4a0a38cfeab4db52a2ddde07f2d0851acbaac

lib/common/socket/evloop/epoll.c.h
sleepybishop: Thank you addressed in 12b8858938581e576acabbf7b89208a9deb4e08d

src/main.c
```diff
@@ -356,9 +364,16 @@ static struct {
...
+         .epoll_bp_usecs = 0,
+         .epoll_bp_budget = 0,
+         .epoll_bp_prefer = false,
+         .epoll_bp_changed = false,
+         .epoll_nonblock = false,
+     }
 };
 
-static __thread size_t thread_index;
+static __thread size_t thread_index = SIZE_MAX;
```
kazuho, line 376: Do we need this change?

src/main.c
```diff
@@ -2550,21 +2571,82 @@ static int open_listener(int domain, int type, int protocol, struct sockaddr *ad
...
+        fprintf(stderr, "dont finish AF_UNIX socket listener\n");
+        return 0;
+    } else if (ss.ss_family == AF_INET) {
+        fprintf(stderr, "will finish AF_INET socket\n");
+    } else if (ss.ss_family == AF_INET6) {
+        fprintf(stderr, "will finish AF_INET6 socket\n");
+    } else {
+        fprintf(stderr, "unknown family: %lu, skipping finish\n", (unsigned long)ss.ss_family);
+        return 0;
+    }
```
kazuho, line 2624: Do we need the `need_finish` flag that we nevertheless make these checks? Maybe we can keep these checks and remove `need_finish`.

src/main.c
```diff
@@ -2550,21 +2571,82 @@ static int open_listener(int domain, int type, int protocol, struct sockaddr *ad
...
 
     /* TCP-specific actions */
-    if (protocol == IPPROTO_TCP) {
+    if (1) {
 #ifdef TCP_DEFER_ACCEPT
         { /* set TCP_DEFER_ACCEPT */
             int flag = 1;
-            if (setsockopt(fd, IPPROTO_TCP, TCP_DEFER_ACCEPT, &flag, sizeof(flag)) != 0)
+            if (setsockopt(fd, IPPROTO_TCP, TCP_DEFER_ACCEPT, &flag, sizeof(flag)) != 0) {
```
kazuho, line 2631: Maebe we need to retain IPPROTO_TCP check for this setsockopt.

t/40server-push-attrs.t
```diff
@@ -69,15 +69,15 @@ unlike $resp, qr{link:\s*\n}, "push-only doesn't appear in the link header";
...
 
 $resp = `nghttp -vn --stat 'https://127.0.0.1:$server->{tls_port}/mruby/?nopush'`;
 unlike $resp, qr{\nid\s*responseEnd\s.*\s/index\.js}is, "index.js is NOT pushed";
-like $resp, qr{link: </index.js>; rel=preload; as=script;}, "Found link header";
+like $resp, qr{link: </index.js>;rel=preload;as=script;}, "Found link header";
 
 $resp = `nghttp -vn --stat 'https://127.0.0.1:$server->{tls_port}/mruby/?x-http2-push-only'`;
 like $resp, qr{\nid\s*responseEnd\s.*\s/index\.js}is, "index.js is pushed";
-unlike $resp, qr{link: </index.js>; rel=preload; as=script;}, "Didn't find link header";
+unlike $resp, qr{link: </index.js>;rel=preload;as=script;}, "Didn't find link header";
```
kazuho, line 80: Do we need these changes?

lib/handler/configurator/busypoll.c
```diff
@@ -0,0 +1,139 @@
+#include "h2o.h"
+#include "h2o/configurator.h"
+
+#include <net/if.h>
+
+static int on_busy_poll_map(h2o_configurator_command_t *cmd, h2o_configurator_context_t *ctx, yoml_t *node)
```
kazuho, line 6: We seem to have config knobs defined in this file and also in src/main.c. Could you elaborate what the distinction between the two are? (FWIW I would not mind everything going into src/main.c, though I generally share the concern that the file is becoming too large)

kazuho (commented):
Thank you for the changes.

I think the PR looks good, but as it it considered experimental and subject to change, I've mainly focused on how the impact of the changes can be isolated.

I'd appreciate trying to build h2o on OS other than Linux (e.g, macOS, FreeBSD) if you happen to have one in hand. I'm happy to point out incompatibilities but it's always easier to ask a computer to find issues (by compling!) rather than going through a human review.

t/40server-push-attrs.t
sleepybishop: This was part of the ci update, reverted in 94b6763a8d1bce0ffd132d04885d46e5666efead and will defer to the other pr.

src/main.c
sleepybishop: No, reverted in cff14365dc3baee042cc638995fad5b1c01868cb

misc/docker-ci/check.mk
```diff
@@ -66,7 +66,7 @@ asan:
 
 # https://clang.llvm.org/docs/SourceBasedCodeCoverage.html
 coverage:
-	docker run $(DOCKER_RUN_OPTS) h2oserver/h2o-ci:ubuntu2204  \
+	docker run $(DOCKER_RUN_OPTS) h2oserver/h2o-ci:ubuntu2404  \
```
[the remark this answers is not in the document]
sleepybishop: agree, reverted

include/h2o.h
```diff
@@ -663,6 +670,11 @@ typedef enum h2o_conn_state {
  * context of the http server.
  */
 struct st_h2o_context_t {
+    /**
+     * thread index/count, used in status handlers for per-thread metrics
+     */
+    int thread_index;
+    int thread_count;
```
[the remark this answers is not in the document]
sleepybishop: No, reverted in cff14365dc3baee042cc638995fad5b1c01868cb

lib/common/socket/evloop/epoll.c.h
```diff
@@ -312,14 +382,32 @@ static void evloop_do_dispose(h2o_evloop_t *_loop)
     struct st_h2o_evloop_epoll_t *loop = (struct st_h2o_evloop_epoll_t *)_loop;
     close(loop->ep);
 }
-h2o_evloop_t *h2o_evloop_create(void)
+
+static h2o_evloop_t *_do_h2o_evloop_create(int flags, uint64_t time_budget, uint64_t packet_budget, uint8_t prefer)
 {
     struct st_h2o_evloop_epoll_t *loop = (struct st_h2o_evloop_epoll_t *)create_evloop(sizeof(*loop));
+    char buf[128];
```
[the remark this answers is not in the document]
sleepybishop: For now yes, it's used a few lines below for the error message

include/h2o/socket/evloop.h
```diff
@@ -22,6 +22,7 @@
 #ifndef h2o__evloop_h
 #define h2o__evloop_h
 
+#include <stdbool.h>
```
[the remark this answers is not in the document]
sleepybishop: Sure, done in 41f88ccef23bdb951a440e742bca46cfd9590932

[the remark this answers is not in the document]
kazuho: Thank you!

lib/common/socket/evloop/epoll.c.h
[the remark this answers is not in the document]
kazuho: Sorry, my question is if we need to move where the variable is defined. It can be at either location, but I am wondering if it is an unintentional change that we want to avoid, to reduce possible conflicts that we might have in the future.

src/main.c
```diff
@@ -2530,6 +2546,11 @@ static int open_listener(int domain, int type, int protocol, struct sockaddr *ad
         }
     }
 #endif
+
+    if (conf.globalconf.bp.nic_count > 0) {
+        h2o_busypoll_bind_interface(fd, iface);
+    }
```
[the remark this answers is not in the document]
sleepybishop: I'de like to but the `SO_BINDTODEVICE` socket option has to be set before the `bind` call in `open_listener`.

lib/core/logconf.c
```diff
@@ -912,7 +920,21 @@ char *h2o_log_request(h2o_logconf_t *logconf, h2o_req_t *req, size_t *len, char
...
+            APPEND_SAFE_STRING(pos, h2o_busypoll_get_iface());
+            break;
+        case ELEMENT_TYPE_BUSYPOLL_NAPI_ID:
+            RESERVE(sizeof(H2O_UINT32_LONGEST_STR) - 1);
+            pos += sprintf(pos, "%" PRIu32, h2o_busypoll_get_napi_id());
+            break;
+        case ELEMENT_TYPE_BUSYPOLL_CPU_IDX:
+            RESERVE(sizeof(H2O_INT16_LONGEST_STR));
+            pos += sprintf(pos, "%" PRIu16, h2o_busypoll_get_cpu_idx());
+            break;
```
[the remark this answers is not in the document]
sleepybishop: i had intended to export `noop` functions in `lib/common/busypoll.c` to avoid guards in multiple places, but i can refactor with guards as needed

lib/common/socket/evloop/epoll.c.h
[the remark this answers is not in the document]
sleepybishop: reverted

src/main.c
sleepybishop: with this logic being split into `finish_open_listener`, `protocol` is no longer available, i could infer it from the fd type being SOCK_STREAM, but currently only tcp sockets are finished. I can try to address together with a refactor eliminating the `need_finish` flag above

lib/handler/configurator/busypoll.c
sleepybishop: I've done some refactoring to minimize loop changes and avoid impacting non-linux platforms. In hindsight it might have been worthwhile making busypoll.h cross platform with a few guards around the cpu set.

kazuho: Thank you for the changes! I see the build succeeding on macOS and linux without ynl.

> In hindsight it might have been worthwhile making busypoll.h cross platform with a few guards around the cpu set.

Yeah I think my comments regarding the direction is based on the assumption that support for busy poll will be an option to the evloop-epoll backend based on if libynl is available, when we have 4 backends (libuv, evloop-epoll, evloop-kqueue, evloop-poll). Also, the busy poll code might change.

Considering these aspects, I have been inclined to not have an abstraction layer, but to keep the code specific as possible.

CMakeLists.txt
```diff
@@ -405,6 +427,7 @@ SET(LIB_SOURCE_FILES
     lib/common/balancer/roundrobin.c
     lib/common/balancer/least_conn.c
     lib/common/absprio.c
+    lib/common/busypoll.c
```
kazuho, line 430: Rather than adding busypoll.c always, I think we might want to follow the precedence of adding feature-specific files only when that feature is available, as well as passing the feature flags using `#define`.

Using mruby support as an example, we do something like below.

In CMakeFiles.txt:
* set `WITH_MRUBY_DEFAULT` based on if dependencies are available,
* set `WITH_MRUBY` based on user input, or adopt `WITH_MRUBY_DEFAULT` if there is no user input, and
* if `WITH_MRUBY` is set, set `-DH2O_USE_MRUBY=1` as the compile flag and add add necessary source files as dependencies.

In the source files, use `#if H2O_USE_MRUBY` to see if the feature is available.

lib/common/busypoll.c
```diff
@@ -0,0 +1,495 @@
+#include "h2o.h"
+
+#if defined(__linux__) && defined(SO_REUSEPORT) && defined(H2O_HAS_YNL_H)
+
```
kazuho, line 4:
```suggestion
```

Assuming we do above, we would no longer need to have these ifs.

lib/common/busypoll.c
```diff
@@ -0,0 +1,495 @@
...
+    return 0;
+}
+
+uint16_t h2o_busypoll_get_cpu_idx(void)
+{
+    return 0;
+}
+
+#endif
+
```
kazuho, line 495:
```suggestion
```
or these stubs.

CMakeLists.txt
kazuho: PS. Maybe a better example is the diff of CMakeLists.txt in https://github.com/h2o/h2o/pull/3201/files.

lib/handler/configurator/busypoll.c
```diff
@@ -0,0 +1,165 @@
...
+            h2o_configurator_scanf(cmd, *st_node, "%zu", &nic->options.suspend_timeout);
+        }
+
+        fprintf(stderr, " ifindex %zu has %zd cpus\n", nic->ifindex, nic->cpu_count);
+    }
+
+    /* setup queues */
+    for (int i = 0; i != nic_count; ++i) {
+        struct busypoll_nic_t *nic = &nic_to_cpu_map->entries[i];
+        if (nic->mode != BP_MODE_OFF) {
```
jdamato-fsly, line 158: FYI, having this check here prevents the map from being populated, which breaks the following config:

```
busy-poll-map:
  interfaces:
    - ifindex: 4
      cpus:
        - 8
        - 9
        - 10
        - 11
      options:
        mode: OFF
 ```

I think it depends if we want to support that configuration or not?

If we do support it, it allows the user to dedicate specific h2o worker threads to specific CPUs that (one would assume) are NUMA local to a NIC -- without enabling busy polling. The user would then create the RSS context and ntuple filters and so on to direct flows to those queues. I could see this being a potentially desirable condition.

If we decide not to support it, then I think, `mode: OFF` can be eliminated because `mode: OFF` results in a fatal:

```
fatal:/home/jdamato/h2o/lib/common/busypoll.c:388:failed to find interface name for napi id: 157
```

Whatever decision is made on this (whether we keep support for mode OFF or not), this might be a good candidate for a test?

File: lib/common/busypoll.c (hunk 20 above)
Comment (sleepybishop) [reply lib/common/busypoll.c]: Thank you for the feedback, I've made the suggested changes in dddeb5a219345615ce8495b3280aec6a5de22201

File: lib/handler/configurator/busypoll.c (hunk 22 above)
Comment (sleepybishop) [reply lib/handler/configurator/busypoll.c]: Thank you, i've removed the `BP_MODE_OFF` guards and added a test for it in 20ea51b2a902fed4f860929fe07faa32c05e7383

File: include/h2o/busypoll.h [hunk 23]
```diff
@@ -0,0 +1,50 @@
...
+
+#if !H2O_USE_EPOLL_BUSYPOLL
+#error "this file may be included only on linux and when ynl support is available"
+#endif
+
+#include <stdint.h>
+
+#define BP_MODE_OFF      0
+#define BP_MODE_SUSPEND  1
+#define BP_MODE_BUSYPOLL 2
```
jdamato-fsly, line 12: seems like this might be a good candidate for an enum instead of defines? but i suppose that is a stylistic choice left to the maintainer

src/main.c
```diff
@@ -2824,8 +2922,15 @@ static int on_config_listen_element(h2o_configurator_command_t *cmd, h2o_configu
                             return -1;
                         }
                     } else {
+                        const char *iface = NULL;
+#if H2O_USE_EPOLL_BUSYPOLL
+                        if (conf.bp.nic_count > 0) {
```
jdamato-fsly, line 2927: FYI this line was formerly:

```
if (conf.bp.nic_count > 0 && conf.bp.nic_to_cpu_map.entries[0].mode != BP_MODE_OFF) {
```

which is what was breaking loopback because loopback does not poll (and is thus `BP_MODE_OFF`) and so `iface` is left set to `NULL` which means that in `open_inet_listener` , `h2o_busypoll_bind_interface` was not actually binding the socket to the device.

this resulted in a reuseport group in the kernel being created that was encompassed all devices for the listener, which then further meant that smaller reuseport groups (which were really just subsets of the larger group) would fail to have their cbpf programs installed, crashing h2o at start.

since this line was changed recently, it should allow h2o to start with the config:

```
busy-poll-map:
  interfaces:
    - ifindex: 1
      cpus:
        - 0
      options:
        mode: OFF
    - ifindex: 6
      cpus:
        - 16
        - 17
        - 18
        - 19
      options:
        gro-flush-timeout: 20000000
        defer-hard-irqs: 100
        suspend-timeout: 80000000
        mode: SUSPEND
```

where `ifindex: 1` is `lo`.

jdamato-fsly: which is to say: perhaps this is a good candidate for a test.

lib/common/busypoll.c
```diff
@@ -0,0 +1,448 @@
...
+    struct netdev_queue_get_req *get_req = NULL;
+    struct netdev_napi_set_req *set_req = NULL;
+    struct ynl_sock *ys;
+    struct ynl_error yerr;
+
+    ys = ynl_sock_create(&ynl_netdev_family, &yerr);
+    if (!ys)
+        h2o_fatal("ynl, failed to create socket: %s\n", yerr.msg);
+
+    if (cfg_ifindex == 1)
```
jdamato-fsly, line 29: i dont really follow this -- is this assuming that the first ifindex is always going to be `lo` and thus no NAPI IDs can be fetched? 

maybe it would be easier, at startup, to use `SIOCSIFFLAGS` to get the flags for every ifindex specified and then know for sure that a particular device is a loopback ?

sleepybishop: thank you, yes the intent was to not fetch napi ids for loopback as that would hit the `h2o_fatal` a few lines down when the `napi id == 0`, ill refactor

src/main.c
sleepybishop: thank you, that makes sense.

src/main.c
```diff
@@ -2553,21 +2587,84 @@ static int open_listener(int domain, int type, int protocol, struct sockaddr *ad
...
+    } else if (ss.ss_family == AF_INET6) {
+        fprintf(stderr, "will finish AF_INET6 socket\n");
+    } else {
+        fprintf(stderr, "unknown family: %lu, skipping finish\n", (unsigned long)ss.ss_family);
+        return 0;
+    }
+
     /* TCP-specific actions */
-    if (protocol == IPPROTO_TCP) {
+    if (1) {
```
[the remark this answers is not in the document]
sleepybishop: the intent was to minimize the size of the diff thinking i could refactor this later to use `SOCK_STREAM` type to infer `IPPROTO_TCP`, but never did, ill go ahead and do that now.

include/h2o/busypoll.h
sleepybishop: thank you, it used to be an enum, but i needed to make the epoll loop implementation aware of the mode to set `max_wait` appropriately and i did not want to include busypoll.h, we can still do that though.

jdamato-fsly: i dont really have a strong preference one way or the other -- just a thing i noticed and commented on. up to you if you want to make it an enum or not

src/main.c
```diff
@@ -3409,6 +3516,8 @@ static int on_config_onoff(h2o_configurator_command_t *cmd, yoml_t *node, int *s
     return 0;
 }
 
+#include "handler/configurator/busypoll.c"
```
kazuho, line 3519: As stated in my comment on CMakeLists.txt, I think we do not want to start adding `lib` as an include directory.

Considering that lib/handler/configurator/busypoll.c defines functions that change the variable `conf`, which is a static variable of src/main.c, I think it might make sense to define those callbacks inside src/main.c.

The alternative would be to do something like what src/ssl.c does. I.e.,
* move lib/hander/configurator/busypoll.c to src/,
* in src/busypoll.c, define a global variable dedicated to busypoll settings,
* add declarations of that global and the functions to src/standalone.h, and
* call them from src/main.c.

include/h2o/busypoll.h
kazuho: Could you elaborate the hesitation behind including busypoll.h from epoll.c.h?

It could very well be the case that something has changed since, but I think we can include files with checking the feature flags. Even epoll.c.h itself is included under the condition of `H2O_USE_EPOLL` being set.

kazuho (commented):
Thank you for the changes! I like how the changes are becoming isolated.

src/main.c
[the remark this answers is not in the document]
sleepybishop: done in b580d622f6743bce79ee35dfad2e28df58a55e50

include/h2o/busypoll.h
sleepybishop: changed back to enum in 13a2745db394da3242e72d767aa697a7a731ce01

lib/common/busypoll.c
sleepybishop: changed to comparing against `lo` in ce30483e827c7ca770d88c5487b62604f17ae4dd.  `SIOCSIFFLAGS` would need a socket to run ioctl on

t/90busypoll.t
```diff
@@ -0,0 +1,207 @@
...
+        chomp $line;
+        $served{$line}++;
+    }
+
+    is(scalar keys %served, $total_threads, "$total_threads thread(s) served all requests");
+    while (my ($served_by, $count) = each %served) {
+        my ($served_iface, $served_napi, $served_cpu, $served_thread) = split ' ', $served_by;
+        if ($served_iface eq $nsim_sv_name) {
+            ok($served_napi > 0, "served by non-zero napi id");
+            ok(abs($count - ($num_attempts/$num_threads)) <= $num_attempts * 0.10, "each queue served share of traffic");
```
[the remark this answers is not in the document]
sleepybishop: added tracking of the stat in 1173fa26e1efeb952edf1e7cd2ea6c74888fe2c3 but i cant get it to increment, can you help me look into it?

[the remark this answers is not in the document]
jdamato-fsly: can you be more specific?

does it not increment when:

- mode = off (no increment is correct here)
- mode = busypoll (no increment is incorrect here)
- mode = suspend (no increment is incorrect here)

also, the config above is incorrect for suspend mode -- suspend mode should enable preferred busy poll to take effect.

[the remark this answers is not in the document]
sleepybishop: right, in `lib/common/socket/evloop/epoll.c.h:444` prefer is set to 1 regardless of the config if the mode is `BP_MODE_SUSPEND`

regarding the stat, i pull `/proc/net/netstat` from the server namespace before and after the curls for each subtest. The busypoll rx packets stat never increments, so it's correct for `mode = off` , but incorrect for `busypoll` and `suspend` modes

lib/common/socket/evloop/epoll.c.h
```diff
@@ -192,6 +233,20 @@ static int update_status(struct st_h2o_evloop_epoll_t *loop)
     return 0;
 }
 
+static void evloop_set_bp(struct st_h2o_evloop_epoll_t *loop, uint64_t usecs, uint64_t budget, uint8_t prefer)
```
kazuho, line 236:
```suggestion
static void set_busypoll(struct st_h2o_evloop_epoll_t *loop, uint64_t usecs, uint64_t budget, uint8_t prefer)
```
"evloop" is not needed for private functions / do not use abbrevs.

src/main.c
```diff
@@ -335,6 +341,17 @@ static struct {
...
+#if H2O_USE_EPOLL_BUSYPOLL
+    struct {
+        size_t nic_count;
+        h2o_busypoll_nic_vector_t nic_to_cpu_map;
+        uint64_t epoll_bp_usecs;
+        uint64_t epoll_bp_budget;
+        uint8_t epoll_bp_prefer : 1;
+        uint8_t epoll_bp_changed : 1;
+        uint8_t epoll_nonblock : 1;
+    } bp;
```
kazuho, line 353:
```suggestion
    } busypoll;
```

kazuho (commented):
Thank you for all the changes. I really like what we have now, all my comments below are nitpicks.

I might try myself to see if we can further reduce the changes to src/main.c, but I am starting to guess I'd fail, as it already looks quite polished up.

CMakeLists.txt
```diff
@@ -156,6 +156,27 @@ ELSE ()
...
+# Find out if system provides libynl
+CMAKE_PUSH_CHECK_STATE()
+FIND_LIBRARY(SYSTEM_YNL_LIB "ynl")
+IF (SYSTEM_YNL_LIB)
+    SET(CMAKE_REQUIRED_LIBRARIES ${SYSTEM_YNL_LIB})
+ENDIF()
+CHECK_C_SOURCE_COMPILES("
+#include <ynl/ynl.h>
+int main(void) {
+return YNL_ERROR_NONE;
```
kazuho, line 168:
```suggestion
    ynl_ntf_free(NULL);
    return 0;
```
We want the linker to fail when libynl is not available. To achieve that, we have to use to a symbol defined in libynl.a rather than using a macro that is resolved by looking solely at the header file.

include/h2o/busypoll.h
```diff
@@ -0,0 +1,48 @@
...
+    enum en_h2o_bp_mode_t mode;
+    cpu_set_t cpu_map;
+    H2O_VECTOR(uint32_t) napi_ids;
+    pthread_mutex_t mutex;
+    struct {
+        size_t gro_flush_timeout;
+        size_t defer_hard_irqs;
+        size_t suspend_timeout;
+    } options;
+    struct busypoll_epoll_params epoll_prm;
```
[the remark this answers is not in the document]
sleepybishop: i've addressed the renames with 0e99b87c706085f4bcda941eb5950dc5ed09249d

include/h2o/socket/evloop.h
```diff
@@ -121,4 +121,7 @@ inline void h2o_timer_link(h2o_evloop_t *loop, uint64_t delay_ticks, h2o_timer_t
     h2o_timerwheel_link_abs(loop->_timeouts, timer, loop->_now_millisec + delay_ticks);
 }
 
+#if H2O_USE_EPOLL_BUSYPOLL
+void h2o_loop_update_bp_params(h2o_evloop_t *loop, uint32_t usecs, uint16_t budget, uint8_t prefer, uint8_t nonblock, uint8_t mode);
+#endif
```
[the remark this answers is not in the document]
sleepybishop: sure, moved and renamed in 1302706258d3695d68b13a9163a5c95ea6914937

lib/common/socket/evloop/epoll.c.h
```diff
@@ -23,22 +23,63 @@
...
 #endif
+#if H2O_USE_EPOLL_BUSYPOLL
+    struct {
+        uint32_t epoll_bp_usecs;
+        uint16_t epoll_bp_budget;
+        uint8_t epoll_bp_prefer : 1;
+        uint8_t epoll_bp_changed : 1;
+        uint8_t epoll_nonblock : 1;
+        uint8_t mode : 2;
+    } bp;
```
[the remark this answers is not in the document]
sleepybishop: Thank you, i've attempted to address the parameter duplication 4dbe87495b0a4fce83fd19d8236ff40cceaf1365

lib/common/socket/evloop/epoll.c.h
```diff
@@ -23,22 +23,63 @@
...
     h2o_io_uring_t io_uring;
 #endif
+#if H2O_USE_EPOLL_BUSYPOLL
+    struct {
+        uint32_t epoll_bp_usecs;
+        uint16_t epoll_bp_budget;
+        uint8_t epoll_bp_prefer : 1;
+        uint8_t epoll_bp_changed : 1;
+        uint8_t epoll_nonblock : 1;
+        uint8_t mode : 2;
```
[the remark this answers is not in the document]
sleepybishop: indeed, also addressed in 4dbe87495b0a4fce83fd19d8236ff40cceaf1365

t/90busypoll.t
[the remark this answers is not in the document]
jdamato-fsly: > regarding the stat, i pull /proc/net/netstat from the server namespace before and after the curls for each subtest. The busypoll rx packets stat never increments, so it's correct for mode = off , but incorrect for busypoll and suspend modes

i tested a simple busy polling app using netdevsim in network namespaces and was able to get the counter to increment. maybe you need to request more data in your test and have the clients send it slower? i would guess its possible that busy polling doesn't even have a chance to run because the FDs are already full of data instantly since its all happening on the same machine at very high rates.

try requesting larger response data and receiving it more slowly?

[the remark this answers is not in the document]
sleepybishop: added more requests with a large payload in 9a6b445f84d1d1ed8301d91d0cd7238d11a9b175. Seems to work in the `mode = suspend` case though kind of inconsistently. I still cannot get the stat to bump for mode `mode = busypoll`, could you share the config that you used?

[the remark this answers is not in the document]
jdamato-fsly: i didnt use h2o, i just use a simple C program to verify that netdevsim + net namespaces would bump the counter

jdamato-fsly:
there's a bug in the loopback setup with this branch, issuing two requests to loopback causes it to fail:

```
thread 0 NAPI id: 0, iface: lo, listener fd: 4
found matching iface in conf map: lo
thread idx 0, has taken cpu: 0, for NIC: lo, via napi: 0 via listener 4

thread 0 NAPI id: 0, iface: lo, listener fd: 4
found matching iface in conf map: lo
no CPU found for thread with NAPI from NIC: lo:0
busy-poll-cpu-map configuration or queue setup is buggy -- exiting h2o now.
```

which shouldn't happen -- the right listener fd is being awoken, so either:
  - multiple threads have the FD for that listener in their epoll loop (they shouldn't)
  - or the code doesn't claim a CPU / thread after the first accept (seems unlikely, since the second request can find no available CPU, meaning the first request triggered the listener to take a CPU and pin itself to that CPU)

sleepybishop:
> possibly worth adding a test which makes multiple requests to both loopback and non-loopback interfaces?

in `t/90busypoll.t`right before sending large requests to the netdevsim interfaces theres a block that sends requests to loopback

```
    for (1..5) {
        my $resp = `ip netns exec nssv curl -ksi http://127.0.0.1:$port`;
    }
```

the same logic runs after the sends to the netdevsim interfaces as well, im not sure why it's not triggering the behavior you are seeing.

jdamato-fsly:
> the same logic runs after the sends to the netdevsim interfaces as well, im not sure why it's not triggering the behavior you are seeing.

Because the test is incorrect.

If you change the test configuration according to the following diff:

```
diff --git a/t/90busypoll.t b/t/90busypoll.t
index 87bad8346..3388450ef 100644
--- a/t/90busypoll.t
+++ b/t/90busypoll.t
@@ -185,7 +185,8 @@ busy-poll-map:
         suspend-timeout: $suspend_timeout
         mode: $mode
     - ifindex: 1
-      cpus: [$total_threads]
+      cpus:
+        - 0
       options:
         mode: OFF
 EOT
```

the problem reproduces:

```
  3.544553   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  3.544583                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0thread 4 NAPI id: 0, iface: lo, listener fd: 11
  3.544789 found matching iface in conf map: lo
  3.544855 thread idx 4, has taken cpu: 0, for NIC: lo, via napi: 0 via listener 11
100   177  100   177    0     0   331k      0 --:--:-- --:--:-- --:--:--  172k
  3.551733   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  3.551763                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0thread 4 NAPI id: 0, iface: lo, listener fd: 11
  3.551931 found matching iface in conf map: lo
  3.552273 no CPU found for thread with NAPI from NIC: lo:0
  3.552277 busy-poll-cpu-map configuration or queue setup is buggy -- exiting h2o now.
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  3.553185 curl: (56) Recv failure: Connection reset by peer
  3.559614   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  3.559631                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  3.559809 curl: (7) Failed to connect to localhost port 32773 after 0 ms: Couldn't connect to server
  3.565928   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  3.565981                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  3.566234 curl: (7) Failed to connect to localhost port 32773 after 0 ms: Couldn't connect to server
  3.572766   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  3.572789                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  3.572957 curl: (7) Failed to connect to localhost port 32773 after 0 ms: Couldn't connect to server
```

sleepybishop:
> If you change the test configuration according to the following diff:

This originated with the change to add the serving cpu index to as a logging directive, the `cpu_claimed` thread local changed from being a boolean to holding the index of the pinned cpu, that introduced an invalid assumption that the cpu indexes started at one. This has been addressed in 3f3a09b8a1e98c021b6b0184bedb40b5d0486ce6

PR #3482, opened 2025-05-08
Fix a crash in the CONNECT handler after timeout with the write of the response header to the client still pending.

jhatala:
Fix a crash in the CONNECT handler after timeout with the write of the response header to the client still pending.

If the write of the response header to the client

```C
h2o_start_response(self->src_req, &self->super);
h2o_send(self->src_req, NULL, 0, H2O_SEND_STATE_IN_PROGRESS);
```
blocks long enough for a timeout to occur and thus for `close_readwrite` to run, then a the `h2o_send` will be still pending even though the payload buffer (`self->tcp.recvbuf_detached->size`) may be empty.

That leads to the write-with-write-pending assert crash:
```
h2o: …/lib/common/socket.c:938: h2o_socket_sendvec: Assertion `sock->_cb.write == NULL' failed.
received fatal signal 6
[92578] …/build/h2o(+0x1804c9)[0x610f8e2b74c9] on_sigfatal at …/src/main.c:3655
[92578] /lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x73704f845330] __sigaction at ??:?
[92578] /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x11c)[0x73704f89eb2c] ?? ??:0
[92578] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x1e)[0x73704f84527e] ?? ??:0
[92578] /lib/x86_64-linux-gnu/libc.so.6(abort+0xdf)[0x73704f8288ff] ?? ??:0
[92578] /lib/x86_64-linux-gnu/libc.so.6(+0x2881b)[0x73704f82881b] ?? ??:0
[92578] /lib/x86_64-linux-gnu/libc.so.6(+0x3b517)[0x73704f83b517] __assert_fail at ??:?
[92578] …/build/h2o(h2o_socket_sendvec+0x490)[0x610f8e1ff840] ?? ??:0
[92578] …/build/h2o(+0x102967)[0x610f8e239967] finalostream_send at …/lib/http1.c:1098 (discriminator 4)
[92578] …/build/h2o(h2o_send+0xe7)[0x610f8e21c1c7] ?? ??:0
[92578] …/build/h2o(h2o_evloop_run+0xa3)[0x610f8e202913] ?? ??:0
[92578] …/build/h2o(+0x183461)[0x610f8e2ba461] run_loop at …/src/main.c:4262
[92578] /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73704f89caa4] pthread_condattr_setpshared at ??:?
[92578] /lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73704f929c3c] __clone at ??:?
```

A fragile repro that today works in my Linux container follows.

The repro first sends a GET request that solicits a response that is sized just so that it fills the socket buffer to the client, then it sends a CONNECT request and the write of the response header blocks.

The socket buffer is prevented from draining by an iptables rule that drops packets, in particular ACKs from the client.

```perl
use strict;
use warnings;
use Time::HiRes qw(sleep);
use Test::More;
use Net::EmptyPort qw(check_port);
use t::Util;

my $origin_port = empty_port();
my $origin = spawn_server(
    argv     => [
        qw(plackup -s Starlet --access-log /dev/null -p), $origin_port, ASSETS_DIR . "/upstream.psgi",
    ],
    is_ready => sub {
        check_port($origin_port);
    },
);

my $server = spawn_h2o(<< "EOT");
hosts:
  default:
    paths:
      "/":
        proxy.connect:
          - "+127.0.0.1"
        proxy.timeout.io: 10
        proxy.connect.emit-proxy-status: ON
      "/p":
        proxy.reverse.url: http://127.0.0.1:$origin_port/
EOT

subtest "repro", sub {
    my $sock = IO::Socket::SSL->new(PeerHost => '127.0.0.1', PeerPort => $server->{tls_port}, SSL_verify_mode => 0) or BAIL_OUT;
    my ($lport, $laddr) = IO::Socket::sockaddr_in($sock->sockname());
    diag "PORT = $lport";
    system("sudo iptables -I INPUT -p tcp -s 127.0.0.1 -d 127.0.0.1 --dport $lport -j DROP");
    my $padlen = 2603863;
    print $sock "GET /p/content?size=$padlen&cl=$padlen HTTP/1.1\r\nHost:localhost\r\n\r\n";
    print $sock "CONNECT 127.0.0.1:$origin_port HTTP/1.1\r\n\r\n";
    sleep 2;
    pass;
};
subtest "ruok", sub {
    my $content = `curl --http1.1 -p -x 127.0.0.1:$server->{port} --silent -v --show-error http://127.0.0.1:$origin_port/echo 2>&1`;
    my @c = $content =~ m{HTTP/[^ ]+ 200\s}mg;
    is @c, 2, "Got two 200 responses";
};

done_testing;
```

### DIFF
--- a/lib/handler/connect.c
+++ b/lib/handler/connect.c
@@ -101,6 +101,10 @@ struct st_connect_generator_t {
      */
     unsigned is_tcp : 1;
     /**
+     * Set just before h2o_start_response, cleared in the generator proceed callback
+     */
+    unsigned resp_header_write_pending : 1;
+    /**
      * TCP- and UDP-specific data
      */
     union {
@@ -254,6 +258,7 @@ static void stop_eyeballs(struct st_connect_generator_t *self)
 
 static void dispose_generator(struct st_connect_generator_t *self)
 {
+    self->resp_header_write_pending = 0;
     stop_eyeballs(self);
     if (self->sock != NULL) {
         h2o_socket_close(self->sock);
@@ -299,6 +304,8 @@ static void close_readwrite(struct st_connect_generator_t *self)
         send_inflight = close_socket(self);
     else if (self->is_tcp)
         send_inflight = self->tcp.recvbuf_detached->size != 0;
+    if (self->resp_header_write_pending)
+        send_inflight = 1;
 
     if (h2o_timer_is_linked(&self->timeout))
         h2o_timer_unlink(&self->timeout);
@@ -640,6 +647,7 @@ static void tcp_on_read(h2o_socket_t *_sock, const char *err)
 static void tcp_on_proceed(h2o_generator_t *_self, h2o_req_t *req)
 {
     struct st_connect_generator_t *self = H2O_STRUCT_FROM_MEMBER(struct st_connect_generator_t, super, _self);
+    self->resp_header_write_pending = 0;
 
     assert(!self->read_closed);
 
@@ -679,6 +687,7 @@ static void tcp_on_connect(h2o_socket_t *_sock, const char *err)
 
     /* build and submit 200 response */
     self->src_req->res.status = 200;
+    self->resp_header_write_pending = 1;
     h2o_start_response(self->src_req, &self->super);
     h2o_send(self->src_req, NULL, 0, H2O_SEND_STATE_IN_PROGRESS);
 }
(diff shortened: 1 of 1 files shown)


jhatala:
That is a good point, I will rethink it from that angle.

PR #3487, opened 2025-05-29
factor out pipe logic

sleepybishop:
This pr subsumes https://github.com/h2o/h2o/pull/3382.

The pipe logic is factored out into `pipe_reader.c` so it could be used outside the request/proxy modules.


include/h2o/pipe_utils.h
```diff
@@ -0,0 +1,39 @@
...
+struct st_h2o_context_t;
+
+typedef struct st_h2o_pipe_reader_t {
+    int fds[2]; /* fd[0] set to -1 unless used */
+    int inflight;
+
+    size_t body_bytes_read, body_bytes_sent;
+} h2o_pipe_reader_t;
+
+int h2o_empty_pipe(int fd);
```
kazuho, line 18: I am not sure if we can have a function that clears any pipes. The pipe reader can clear it because it knows how data is written to the pipe.

Considering that, IMO it would make sense to:
* turn this function into a static function inside .c, and
* rename the added files to pipe_reader.[ch], as they would then deal only with pipe readers.

include/h2o/pipe_utils.h
```diff
@@ -0,0 +1,39 @@
...
+
+void h2o_pipe_reader_send(struct st_h2o_req_t *req, h2o_pipe_reader_t *pipe_reader, enum h2o_send_state send_state);
```
kazuho, line 20: I am using this declaration as an example, but in general, I would not mind calling the 2nd argument "reader", because maybe that is obvious. Alongside the inclusion of h2o.h (see my comment above), the declaration would then look like:
```suggestion
void h2o_pipe_reader_send(h2o_req_t *req, h2o_pipe_reader_t *reader, h2o_send_state_t send_state);
```

include/h2o/pipe_utils.h
```diff
@@ -0,0 +1,39 @@
...
+
+int h2o_pipe_reader_start(struct st_h2o_context_t *ctx, h2o_pipe_reader_t *pipe_reader);
+
+void h2o_pipe_reader_dispose(struct st_h2o_context_t *ctx, h2o_pipe_reader_t *pipe_reader);
+
+int h2o_pipe_reader_is_empty(h2o_pipe_reader_t *pipe_reader);
+
+void h2o_pipe_reader_update(h2o_pipe_reader_t *pipe_reader, size_t read_body_bytes);
+
+static inline bool h2o_pipe_reader_is_initialized(h2o_pipe_reader_t *pipe_reader)
```
kazuho, line 34: I think the name is incorrect, as this function returns false when the file descriptor is set to -1, even though the "init" function initializes the value to -1.

Maybe something like "is_in_use"?

kazuho (commented):
@sleepybishop Thank you, below are my reviews regarding the design.

I haven't really read through the *.c files, but I believe they are mostly fine, as this PR is basically a refactor.

include/h2o/pipe_utils.h
sleepybishop: Thank you, fixed in e34e5903b538548260b59ca2db646e6d6720734c

sleepybishop:
Thank you for the refactor, LGTM!

PR #3488, opened 2025-06-30
[http1client] do not opt-in the pipe reader if the client is using ssl

deweerdt:
the optimization doesn't make sense in this case since it would result in encrypted bytes being fed to the pipe.


deweerdt:
I feel like I worded the comment poorly. This isn't a perf issue but a correctness one: the code currently writes encrypted bytes in the pipe that are then spliced as is by the H1 frontend.

PR #3492, opened 2025-07-17
Updated H2O to support the newly added KEMs in PicoTLS.

rsith71:
Updating H2O to use the new ML_KEM algorithms provided by PicoTLS.

I wrote up a automation test. Though I could not get it to run on the Docker CI container since it did not have OpenSSL 3.5+. Therefore I ran the tests on my development machine.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1329,6 +1329,16 @@ static const char *listener_setup_ssl_picotls(struct listener_config_t *listener
 #else
             &ptls_minicrypto_x25519,
 #endif
+#if PTLS_OPENSSL_HAVE_X25519MLKEM768
+            &ptls_openssl_x25519mlkem768,
+#endif
+#if PTLS_OPENSSL_HAVE_MLKEM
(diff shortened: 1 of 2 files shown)


rsith71:
@vparla@cisco.com -- Vinny here is the change for H2O to use ML-KEM algorithms.

rsith71:
@kazuho @vparla 

Question: If you want to configure an H2O server to only allow a specific Group/KEM or a list of Group/KEMs it is configured via the config file's SSL section under the parameter: key-exchange-tls1.3
and the values should be from here: https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8

Is that correct?

rsith71:
I can see this waiting. It just means that it is a configuration change in your H2O.conf file right now. 

By the way I was doing some testing and saw this issue. https://github.com/h2o/h2o/issues/3505

I just created it and didn't have time to look into it. Let me know if you have seen this before or need more information.

Thank you.

PR #3580, opened 2026-04-18
Trae/solo agent h gnbt h

sourcetf:


sourcetf:
modified for openbsd,may not work on linux

PR #3582, opened 2026-04-26
masque: use RFC 9298 well-known path for CONNECT-UDP handler

aritrbas:
- Update test configuration and all test URLs to use the RFC-mandated path `/.well-known/masque/udp/` for `CONNECT-UDP` instead of the non-standard `/rfc9298`.
- Improve documentation to clearly recommend registering `proxy.connect-udp` at `/.well-known/masque/udp` with an example config snippet.
- **No changes in handler logic; only configuration and documentation are updated** so users are less likely to misconfigure their deployments and ensure interoperability.


kazuho:
Thank you for the PR! Merged.

PR #3588, opened 2026-05-11
Implement port range ACL

max-b:
Adds support for port ranges in proxy.connect ACL list.


max-b:
Looks like I lost a thing or two in translation from the other repo, I'll fix these

t/50connect-fin-only.t
```diff
@@ -20,7 +20,7 @@ hosts:
     paths:
       "/":
         proxy.connect:
-          - "+127.0.0.1:*"
```
max-b, line 23: This is the one backwards incompatible change, but I wasn't sure that this was a desired format and it's not actually [specified](https://h2o.examp1e.net/configure/proxy_directives.html#proxy.connect)

> Each element of the access control list starts with either + or - followed by an wildcard (*) or an IP address with an optional netmask and an optional port number.

PR #3591, opened 2026-05-26
[HTTP/3] decoder support for QPACK dynamic table

kazuho:
This PR does the following:
* Wires and enables support for QPACK dynamic tables on the decoding side into the HTTP/3 request path.
* Sets the default table size in both directions (i.e., encoder and decoder) to 16KB (though note the encoder is not used (yet) regardless of the value set).
* Fixes bugs in the dynamic table decoding support that has existed long inside qpack.[ch] but never was exercised outside unit tests.
* Updates qif/unit coverage for the adjusted decoder API
* The decoder table size can be set using the `qpack-decoder-table-capacity` attribute within `listen` -> `quic`.


lib/common/http3client.c
```diff
@@ -539,9 +539,10 @@ static quicly_error_t handle_input_expect_headers(struct st_h2o_http3client_req_
             return 0;
         }
     }
+    uint64_t qpack_blocked_ref;
     if ((ret = h2o_qpack_parse_response(req->super.pool, req->conn->super.qpack.dec, req->quic->stream_id, &status, &headers,
-                                        &datagram_flow_id, header_ack, &header_ack_len, frame.payload, frame.length, err_desc)) !=
-        0) {
+                                        &datagram_flow_id, &qpack_blocked_ref, header_ack, &header_ack_len, frame.payload,
```
kazuho: Maybe make `qpack_block_ref` an optional out parameter that is required only when dynamic table is activated?

include/h2o/http3_common.h
```diff
@@ -288,11 +288,16 @@ typedef struct st_h2o_http3_qpack_context_t {
...
+    /**
+     * Table capacity we advertise to the peer for its encoder.
+     */
+    uint32_t decoder_table_capacity;
 } h2o_http3_qpack_context_t;
 
 typedef struct st_h2o_http3_conn_callbacks_t {
     h2o_quic_conn_callbacks_t super;
     void (*handle_control_stream_frame)(h2o_http3_conn_t *conn, uint64_t type, const uint8_t *payload, size_t len);
+    void (*handle_qpack_unblocked_streams)(h2o_http3_conn_t *conn, uint64_t insert_count);
```
kazuho: `qpack_unblock_streams`?

lib/http3/server.c
```diff
@@ -122,6 +122,10 @@ struct st_h2o_http3_server_conn_t {
          * holds streams in REQ_PENDING state or RECV_BODY_POST_BLOCK state (that is using streaming; i.e., write_req.cb != NULL).
          */
         h2o_linklist_t pending;
+        /**
+         * holds streams whose header blocks are blocked by QPACK dynamic table references.
```
kazuho: The doc-comment should clarify, like is done in those of the syblings, streams of which state is to be linked. The states cannot overlap, correct?

lib/http3/server.c
```diff
@@ -245,6 +249,11 @@ struct st_h2o_http3_server_stream_t {
      * indicates if the request is in streaming mode
      */
     uint8_t req_streaming : 1;
+    uint64_t qpack_blocked_ref;
```
kazuho: This location looks odd, and we need a doc-comment.

lib/http3/server.c
```diff
@@ -479,13 +488,26 @@ static void pre_dispose_request(struct st_h2o_http3_server_stream_t *stream)
...
 
+static void cancel_qpack_blocked_stream(struct st_h2o_http3_server_stream_t *stream)
+{
+    if (stream->qpack_blocked_ref == 0)
+        return;
+    if (h2o_linklist_is_linked(&stream->link))
+        h2o_linklist_unlink(&stream->link);
+    stream->qpack_blocked_ref = 0;
+    h2o_qpack_decoder_update_num_blocked(get_conn(stream)->h3.qpack.dec, -1);
+    h2o_http3_send_qpack_stream_cancel(&get_conn(stream)->h3, stream->quic->stream_id);
```
kazuho: The first line calls the qpack decoder directly, and the second line does so indirectly. Is there a reason to define them as separate actions? It'd be easy to miss the other...

Also, is it correct to send Stream Cancellation only if the stream is blocked?

lib/http3/server.c
```diff
@@ -479,13 +488,26 @@ static void pre_dispose_request(struct st_h2o_http3_server_stream_t *stream)
...
+}
+
 static void set_state(struct st_h2o_http3_server_stream_t *stream, enum h2o_http3_server_stream_state state, int in_generator)
 {
     struct st_h2o_http3_server_conn_t *conn = get_conn(stream);
     enum h2o_http3_server_stream_state old_state = stream->state;
 
     H2O_PROBE_CONN(H3S_STREAM_SET_STATE, &conn->super, stream->quic->stream_id, (unsigned)state);
 
+    assert(stream->qpack_blocked_ref == 0);
```
kazuho: You should clarify why, either as `assert(... && "explanation")` or as a comment.

lib/http3/server.c
```diff
@@ -792,6 +821,7 @@ void on_stream_destroy(quicly_stream_t *qs, quicly_error_t err)
 
     req_scheduler_deactivate(&conn->scheduler.reqs, &stream->scheduler);
 
+    assert(stream->qpack_blocked_ref == 0);
```
kazuho: I do not understand the necessity of this line, when the comment right above states that the stream must be in CLOSE_WAIT, and when we have a guard that prevents transitioning to CLOSE_WAIT while the stream is qpack-blocked. _If_ there are concerns, wouldn't it be better to check state at the beginning of this function?

lib/http3/server.c
```diff
@@ -1399,13 +1433,25 @@ static quicly_error_t handle_input_expect_headers(struct st_h2o_http3_server_str
...
     if ((ret = h2o_qpack_parse_request(&stream->req.pool, get_conn(stream)->h3.qpack.dec, stream->quic->stream_id,
                                        &stream->req.input.method, &stream->req.input.scheme, &stream->req.input.authority,
                                        &stream->req.input.path, &stream->req.upgrade, &stream->req.headers, &header_exists_map,
                                        &stream->req.content_length, &expect, NULL /* TODO cache-digests */, &datagram_flow_id_field,
-                                       header_ack, &header_ack_len, frame.payload, frame.length, err_desc)) != 0 &&
-        ret != H2O_HTTP2_ERROR_INVALID_HEADER_CHAR)
+                                       &qpack_blocked_ref, header_ack, &header_ack_len, frame.payload, frame.length, err_desc)) != 0 &&
+        ret != H2O_HTTP2_ERROR_INVALID_HEADER_CHAR) {
+        if (ret == H2O_HTTP3_ERROR_INCOMPLETE) {
+            assert(qpack_blocked_ref != 0);
```
kazuho: Is there a reason to use an error code _and_ an out parameter rather than calling it a success with the out parameter set to a non-zero value?

lib/http3/server.c
```diff
@@ -1399,13 +1433,25 @@ static quicly_error_t handle_input_expect_headers(struct st_h2o_http3_server_str
...
+            stream->recvbuf.handle_input = handle_input_expect_headers;
```
kazuho: Is there a reason to re-assign?

lib/http3/server.c
```diff
@@ -1399,13 +1433,25 @@ static quicly_error_t handle_input_expect_headers(struct st_h2o_http3_server_str
...
+            stream->qpack_blocked_ref = qpack_blocked_ref;
```
kazuho: Is there a reason to use a local variable rather than passing `&stream->qpack_blocked_ref` to the parse function?

lib/http3/server.c
```diff
@@ -1399,13 +1433,25 @@ static quicly_error_t handle_input_expect_headers(struct st_h2o_http3_server_str
...
+            stream->qpack_blocked_ever = 1;
+            h2o_qpack_decoder_update_num_blocked(conn->h3.qpack.dec, 1);
+            h2o_linklist_insert(&conn->delayed_streams.qpack_blocked, &stream->link);
+            *src = frame_start;
```
kazuho: Is it our convention to keep the original value somewhere and restore it, rather than to use a local variable that gets updated and, upon success, set that local variable as the output?

include/h2o/http3_common.h
```diff
@@ -463,7 +463,7 @@ int h2o_http3_handle_settings_frame(h2o_http3_conn_t *conn, const uint8_t *paylo
 /**
  *
  */
-void h2o_http3_send_qpack_stream_cancel(h2o_http3_conn_t *conn, quicly_stream_id_t stream_id);
+void h2o_http3_send_qpack_stream_cancel(h2o_http3_conn_t *conn, quicly_stream_id_t stream_id, int is_blocked);
```
kazuho: We need to update the name too? This function is no longer just sending stream cancellation.

include/h2o/qpack.h
```diff
@@ -51,7 +51,9 @@ size_t h2o_qpack_decoder_send_state_sync(h2o_qpack_decoder_t *qpack, uint8_t *ou
 size_t h2o_qpack_decoder_send_stream_cancel(h2o_qpack_decoder_t *qpack, uint8_t *outbuf, int64_t stream_id);
 
 /**
- * Parses a QPACK request. The input should be the *payload* of the HTTP/3 HEADERS frame.
+ * Parses a QPACK request. The input should be the *payload* of the HTTP/3 HEADERS frame. If `blocked_ref` is non-NULL and the
+ * field section is blocked, returns success with `*blocked_ref` set to the Required Insert Count. If `blocked_ref` is NULL, the
+ * decoder treats a blocked field section as a decompression error.
```
kazuho: Isn't it a contract violation to pass NULL when dynamic tables are used? I think we hard fail in other cases under the same condition.

lib/http3/qpack.c
```diff
@@ -790,10 +764,12 @@ static int parse_decode_context(h2o_qpack_decoder_t *qpack, struct st_h2o_qpack_
...
     /* is the stream blocked? */
     if (ctx->req_insert_count >= qpack_table_total_inserts(&qpack->table)) {
-        if (qpack->blocked_streams.list.size >= qpack->max_blocked)
+        if (qpack->num_blocked >= qpack->max_blocked)
             return H2O_HTTP3_ERROR_QPACK_DECOMPRESSION_FAILED;
-        decoder_link_blocked(qpack, stream_id, ctx->req_insert_count);
-        return H2O_HTTP3_ERROR_INCOMPLETE;
+        if (blocked_ref == NULL)
+            return H2O_HTTP3_ERROR_QPACK_DECOMPRESSION_FAILED;
+        *blocked_ref = ctx->req_insert_count;
```
kazuho: Am I correct to understand that `parse_decode_context` returns the context including `req_insert_count`, _and also return the same value in `*blocked_ref`? Why?

lib/http3/server.c
```diff
@@ -123,7 +123,9 @@ struct st_h2o_http3_server_conn_t {
          */
         h2o_linklist_t pending;
         /**
-         * holds streams whose header blocks are blocked by QPACK dynamic table references.
+         * holds streams in RECV_HEADERS state whose header blocks are blocked by QPACK dynamic table references.
+         * This use of st_h2o_http3_server_stream_t::link cannot overlap with other delayed_streams lists because
+         * QPACK-blocked streams have not advanced beyond request header parsing.
```
kazuho: Please drop the 2nd sentence, we do not have such "excuse" written out in siblings.

lib/http3/server.c
```diff
@@ -488,15 +493,15 @@ static void pre_dispose_request(struct st_h2o_http3_server_stream_t *stream)
         --get_conn(stream)->num_streams_tunnelling;
 }
 
-static void cancel_qpack_blocked_stream(struct st_h2o_http3_server_stream_t *stream)
+static void cancel_qpack_decoder_stream(struct st_h2o_http3_server_stream_t *stream)
 {
-    if (stream->qpack_blocked_ref == 0)
+    int is_blocked = stream->qpack_blocked_ref != 0;
+    if (!is_blocked && stream->state != H2O_HTTP3_SERVER_STREAM_STATE_RECV_HEADERS)
```
kazuho: I know you checked RFC 9204. Clarify why this is correct.

lib/http3/server.c
kazuho: Maybe you forgot to resolve this?

lib/http3/server.c
```diff
@@ -501,7 +501,9 @@ static void cancel_qpack_decoder_stream(struct st_h2o_http3_server_stream_t *str
     if (h2o_linklist_is_linked(&stream->link))
         h2o_linklist_unlink(&stream->link);
     stream->qpack_blocked_ref = 0;
-    h2o_http3_send_qpack_stream_cancel(&get_conn(stream)->h3, stream->quic->stream_id, is_blocked);
+    if (is_blocked)
+        h2o_qpack_decoder_update_num_blocked(get_conn(stream)->h3.qpack.dec, -1);
+    h2o_http3_send_qpack_stream_cancel(&get_conn(stream)->h3, stream->quic->stream_id);
```
kazuho: Could you clarify why you reverted the change we agreed?

lib/http3/common.c
```diff
@@ -1263,7 +1263,7 @@ void h2o_http3_dispose_conn(h2o_http3_conn_t *conn)
 
 static uint64_t calc_max_blocked_streams(h2o_http3_conn_t *conn)
 {
-    if (conn->qpack.ctx->decoder_table_capacity == 0)
+    if (conn->qpack.ctx->decoder_table_capacity == 0 || conn->super.callbacks->qpack_unblock_streams == NULL)
```
kazuho: This contradicts with the pattern added in the same commit stating that "if the decoder allows blocked streams, `blocked_ref` must be non-NULL".

lib/http3/qpack.c
```diff
@@ -762,16 +760,21 @@ static int parse_decode_context(h2o_qpack_decoder_t *qpack, struct st_h2o_qpack_
...
+{
+    if (blocked_ref != NULL)
+        *blocked_ref = 0;
+    if (ctx->req_insert_count < qpack_table_total_inserts(&qpack->table))
+        return 0;
+    if (qpack->num_blocked >= qpack->max_blocked)
+        return H2O_HTTP3_ERROR_QPACK_DECOMPRESSION_FAILED;
+    if (blocked_ref == NULL)
+        return H2O_HTTP3_ERROR_QPACK_DECOMPRESSION_FAILED;
+    *blocked_ref = ctx->req_insert_count;
```
kazuho, line 789: Is it necessary to have this function? The logic that does not touch blocked_ref could be inside `parse_decode_context`, and the rest could be in the caller, integrated with other handling of blocked_ref?

lib/http3/server.c
```diff
@@ -496,6 +496,8 @@ static void cancel_qpack_decoder_stream(struct st_h2o_http3_server_stream_t *str
     int is_blocked = stream->qpack_blocked_ref != 0;
     if (!is_blocked && stream->state != H2O_HTTP3_SERVER_STREAM_STATE_RECV_HEADERS)
         return;
+    /* RFC 9204 requires Stream Cancellation when a stream is reset before all encoded field sections have been processed, or when
+     * reading is abandoned. If `is_blocked` is set, release our local blocked-stream budget as well. */
```
kazuho: Did you actually read the RFC to think that the above if is correct?

PR #3596, opened 2026-05-30
Prevent Undefined behaviour with nbits and pbits.

loganaden:
(Thanks to Frank Denis)

Signed-off-by: Codarren Velvindron


lib/http2/cache_digests.c
```diff
@@ -75,6 +75,8 @@ static void load_digest(h2o_cache_digests_t **digests, const char *gcs_base64, s
     /* decode nbits and pbits */
     if (golombset_decode_bits(&ctx, 5, &nbits) != 0 || golombset_decode_bits(&ctx, 5, &pbits) != 0)
         goto Exit;
+    if ((nbits + pbits) > 63)
+	goto Exit;
```
kazuho, line 79: Thank you for the PR. As both `n` and `p` are decodes as 5-bits (see the code above), the addition remains below 63.

PR #3603, opened 2026-06-15
Enable QPACK fill-till-full for responses

kazuho:
Enable HTTP/3 responses to use the QPACK dynamic table in a fill-till-full mode. The server response path now passes an encoder buffer to the QPACK flattener and writes any produced encoder-stream instructions to the egress QPACK encoder stream.

The insertion policy remains deliberately simple: insert while an entry fits, never evict. It keeps the existing value-length threshold for names that are already referenceable, skips `etag`, and skips fields marked `dont_compress` / never-index candidates.

The design doc of the encoder can be found at: https://gist.github.com/kazuho/fb36ec78ccd062f341d235e0211330c0.


lib/http3/common.c
```diff
@@ -1478,6 +1478,15 @@ void h2o_http3_send_qpack_header_ack(h2o_http3_conn_t *conn, const void *bytes,
     H2O_HTTP3_CHECK_SUCCESS(quicly_stream_sync_sendbuf(stream->quic, 1));
 }
 
+void h2o_http3_send_qpack_encoder_instructions(h2o_http3_conn_t *conn, const void *bytes, size_t len)
+{
+    struct st_h2o_http3_egress_unistream_t *stream = conn->_control_streams.egress.qpack_encoder;
+
+    assert(stream != NULL);
+    h2o_buffer_append(&stream->sendbuf, bytes, len);
+    H2O_HTTP3_CHECK_SUCCESS(quicly_stream_sync_sendbuf(stream->quic, 1));
```
kazuho: Do we need to add this dedicated function when the code is so short and have only one caller?

lib/http3/qpack.c
```diff
@@ -1123,7 +1123,21 @@ static void flatten_without_nameref(struct st_h2o_qpack_flatten_context_t *ctx,
     flatten_string(&ctx->headers_buf, value.base, value.len, 7, dont_compress);
 }
 
-static void do_flatten_header(struct st_h2o_qpack_flatten_context_t *ctx, int32_t static_index, int is_exact, int likely_to_repeat,
+static int should_insert_dynamic(const h2o_iovec_t *name, h2o_header_flags_t flags)
+{
+    if (flags.dont_compress)
```
kazuho: I'm not sure the flag is meant to disable use of dynamic tables. Please consult RFC 9204 to find out why fields that have this flag set need special care.

lib/http3/qpack.c
```diff
@@ -1123,7 +1123,21 @@ static void flatten_without_nameref(struct st_h2o_qpack_flatten_context_t *ctx,
...
+        return 0;
+    if (h2o_iovec_is_token(name)) {
+        const h2o_token_t *token = H2O_STRUCT_FROM_MEMBER(h2o_token_t, buf, name);
+        if (token->flags.dont_compress || token == H2O_TOKEN_ETAG)
+            return 0;
+    } else if (h2o_memis(name->base, name->len, H2O_STRLIT("etag"))) {
```
kazuho: We do not need this, as the field names are always normalized (and if you believe the contrary, you should be checking uppercase too and have the logic to lowercase them).

lib/http3/qpack.c
```diff
@@ -1143,16 +1157,16 @@ static void do_flatten_header(struct st_h2o_qpack_flatten_context_t *ctx, int32_
             flatten_dynamic_indexed(ctx, dynamic_index);
             return;
         }
-        /* emit to encoder buf and dynamic index?
-         * At the moment the strategy is dumb; we emit encoder stream data until the table becomes full. Never triggers eviction.
-         */
-        if (likely_to_repeat && ctx->encoder_buf != NULL && ((static_index < 0 && dynamic_index < 0) || value.len >= 8) &&
+        /* Fill the dynamic table while there is room. Never evicts. */
```
kazuho:
```suggestion
        /* Fill the dynamic table while there is room. */
```

lib/http3/qpack.c
```diff
@@ -1143,16 +1157,16 @@ static void do_flatten_header(struct st_h2o_qpack_flatten_context_t *ctx, int32_
...
+        if (ctx->encoder_buf != NULL && should_insert_dynamic(name, flags) &&
+            ((static_index < 0 && dynamic_index < 0) || value.len >= 8) &&
```
kazuho, line 1147: This could be an issue with the original code that never gets exercised, but I do not follow `|| value.len >=8`.

Assuming there's room in the dynamic table, IMO we want to insert:
* If `value.len >= 8`, always.
* If `value.len < 8`, only if a name cannot be indexed (neither static or dynamic).

lib/http3/qpack.c
```diff
@@ -1164,8 +1178,9 @@ static void do_flatten_header(struct st_h2o_qpack_flatten_context_t *ctx, int32_
             } else {
                 added =
                     h2o_mem_alloc_shared(NULL, offsetof(struct st_h2o_qpack_header_t, value) + name->len + 1 + value.len + 1, NULL);
+                added->value_len = value.len;
                 added->name = &added->_name_buf;
-                added->_name_buf = h2o_iovec_init(added->value + added->value_len + 1, name->len);
+                added->_name_buf = h2o_iovec_init(added->value + value.len + 1, name->len);
```
kazuho: Is this a bug fix or an unneeded change?

lib/http3/qpack.c
```diff
@@ -1193,16 +1208,15 @@ static void do_flatten_header(struct st_h2o_qpack_flatten_context_t *ctx, int32_
...
 {
     int32_t static_index = -1;
-    int is_exact = 0, likely_to_repeat = 0;
+    int is_exact = 0;
 
     /* obtain static index if possible */
     if (h2o_iovec_is_token(header->name)) {
         const h2o_token_t *token = H2O_STRUCT_FROM_MEMBER(h2o_token_t, buf, header->name);
         static_index = h2o_qpack_lookup_static[token - h2o__tokens](header->value, &is_exact);
-        likely_to_repeat = token->flags.likely_to_repeat;
```
kazuho, line 1195: Now that we're deleting this reference, do we still use the flag elsewhere? If not, we should remove it (and as the table is generated, the generator needs to be fixed).

t/00unit/lib/http3/qpack.c
```diff
@@ -122,6 +122,56 @@ static void test_simple(void)
...
+    h2o_qpack_section_stats_t stats = {0};
+    const char *err_desc = NULL;
+
+    h2o_mem_init_pool(&pool);
+
+    h2o_add_header(&pool, &headers, H2O_TOKEN_CONTENT_TYPE, NULL, H2O_STRLIT("text/foo"));
+    h2o_add_header_by_str(&pool, &headers, H2O_STRLIT("x-fill"), 0, NULL, H2O_STRLIT("y"));
+    h2o_add_header(&pool, &headers, H2O_TOKEN_CONTENT_TYPE, NULL, H2O_STRLIT("text/bar"));
+    h2o_add_header(&pool, &headers, H2O_TOKEN_ETAG, NULL, H2O_STRLIT("\"abc\""));
+    h2o_add_header(&pool, &headers, H2O_TOKEN_SET_COOKIE, NULL, H2O_STRLIT("sid=secret"));
```
kazuho: Please clarify the intent behind these names and values either as a comment or in the used literal values themselves. Otherwise it is hard to understand what is tested and not.

kazuho (commented):
Thank you for the PR.

Could you clarify how we handle max_blocked?

Also, please consult my review comments inline.

lib/http3/common.c
```diff
@@ -1485,12 +1485,10 @@ void h2o_http3_qpack_cancel_stream(h2o_http3_conn_t *conn, quicly_stream_id_t st
     H2O_HTTP3_CHECK_SUCCESS(quicly_stream_sync_sendbuf(stream->quic, 1) == 0);
 }
 
-void h2o_http3_send_qpack_header_ack(h2o_http3_conn_t *conn, const void *bytes, size_t len)
+void h2o_http3_write_unistream(struct st_h2o_http3_egress_unistream_t *stream, const void *data, size_t len)
```
kazuho:
```suggestion
void h2o_http3_write_unistream(struct st_h2o_http3_egress_unistream_t *stream, const void *bytes, size_t len)
```
We're sending data as bytes, so what's the purpose of the change?

lib/http3/common.c
```diff
@@ -1485,12 +1485,10 @@ void h2o_http3_qpack_cancel_stream(h2o_http3_conn_t *conn, quicly_stream_id_t st
...
 {
-    struct st_h2o_http3_egress_unistream_t *stream = conn->_control_streams.egress.qpack_decoder;
-
     assert(stream != NULL);
-    h2o_buffer_append(&stream->sendbuf, bytes, len);
+    h2o_buffer_append(&stream->sendbuf, data, len);
```
kazuho:
```suggestion
    h2o_buffer_append(&stream->sendbuf, bytes, len);
```

lib/http3/qpack.c
```diff
@@ -1116,16 +1116,40 @@ static void flatten_without_nameref(struct st_h2o_qpack_flatten_context_t *ctx,
...
+}
+
+static int should_insert_dynamic(const h2o_iovec_t *name, int dont_compress)
+{
+    if (dont_compress)
+        return 0;
+    if (h2o_iovec_is_token(name) && H2O_STRUCT_FROM_MEMBER(h2o_token_t, buf, name) == H2O_TOKEN_ETAG)
+        return 0;
+    return 1;
+}
```
kazuho: Is there a reason to define these functions (calc_effective_dont_compress and should_insert_dynamic) as dedicated functions?

Separately from that, please add a comment stating that the handling of dont_compress matches what we have for HPACK.

misc/tokens.pl
```diff
@@ -59,11 +59,11 @@
     chomp $line;
     last if $line eq '';
     next if $line =~ /^#/;
-    my ($hpack_index, $proxy_should_drop_for_req, $proxy_should_drop_for_res, $is_init_header_special, $is_hpack_special, $copy_for_push_request, $dont_compress, $likely_to_repeat, $name, $value) =
+    my ($hpack_index, $proxy_should_drop_for_req, $proxy_should_drop_for_res, $is_init_header_special, $is_hpack_special, $copy_for_push_request, $dont_compress, $unused_repeat_hint, $name, $value) =
```
kazuho: When asked to remove, please remove the field in the table rather than just removing from the struct and while keeping the field in the table.

t/00unit/lib/http3/qpack.c
```diff
@@ -122,6 +122,133 @@ static void test_simple(void)
...
+    h2o_headers_t headers = {NULL};
+    h2o_byte_vector_t enc_stream = {NULL};
+    h2o_qpack_section_stats_t stats = {0};
+    const char *err_desc = NULL;
+
+    h2o_mem_init_pool(&pool);
+
+    h2o_add_header(&pool, &headers, H2O_TOKEN_CONTENT_TYPE, NULL, H2O_STRLIT("text/fill-a"));   /* static name */
+    h2o_add_header_by_str(&pool, &headers, H2O_STRLIT("x-fill"), 0, NULL, H2O_STRLIT("y"));      /* no name ref */
+    h2o_add_header(&pool, &headers, H2O_TOKEN_CONTENT_TYPE, NULL, H2O_STRLIT("text/fill-b"));   /* dynamic name */
```
kazuho: I do not follow the rationale behind the choice (or that the comments are good).

I'm not sure if the list below is exhaustive, but don't we want to check at least:
* name-and-value match against the static table or the dynamic table
* name-only match against the static table or the dynamic table that causes a name-value insert
* name-only match against the static table or the dynamic table that does not cause an insert due to the value being too short
* non-match against neither table that causes an name-value insert regardless of the value being short or long

kazuho:
Subsumed by #3606.

PR #3606, opened 2026-06-19
[HTTP/3] Use QPACK dynamic table for sending, managed by churn cost estimator

kazuho:
Subsuming #3603, which lets responses use the QPACK dynamic table in **fill-till-full** mode (insert while an entry fits, never evict), this PR implements the full-shape: insert into the dynamic table based on the expected value of replacing an existing entry, so a small table keeps tracking the working set instead of freezing. This implements **"strategy 16"** from the encoder design doc: https://gist.github.com/kazuho/fb36ec78ccd062f341d235e0211330c0.

The encoder does all of this on its own; the decoder is unchanged and unaware. Configurable per endpoint, default on.

In addition, this PR wires h2o-httpclient's use of dynamic table on the receiving side to run end-to-end tests between h2o and h2o-httpclient.

### How it works
- **Shadow cache** — a small, fixed-size, encoder-only 4-way set-associative cache (~½ the table size) holding a decayed frequency for *non-resident* `(name,value)` pairs. Never sent on the wire.
- **Score** = `decayed_freq × bytes_saved / entry_size` — value per byte of table held, so it favors pairs that save more. Frequency decays by a per-section growth factor and *grows-and-freezes* near float range rather than rescaling, so refinement naturally winds down after ~10K+ sections — long past the point the table has converged.
- **Hysteretic swap** — once full, a non-resident pair that has gathered repeat evidence *and* beats the weakest resident by a 2× margin triggers a swap. Residents dug past in the FIFO that the candidate doesn't beat are preserved with a QPACK **Duplicate** (moved to the tail), so hot-but-old entries survive.
- Per-field cost stays O(1) (cached in-flight reference floor; minimum resident score recomputed lazily during the table walk the encoder already performs), so refinement adds no unbounded per-field work.

### Why this approach

The core insight is that, before the table is full, inserting a field does not displace any existing dynamic-table entry. If that field later proves weak, nothing has to be undone; it can remain until the table fills and then lose naturally to better candidates. But once the table is full, insertions turn into replacements that change the resident set. If the replacement was wrong, restoring the previous resident set requires additional QPACK encoder instructions to be sent. Therefore the policy fills eagerly first, then gates replacements once mistakes become costly to reverse.

- **Fast startup.** Compression is immediate — the table populates from the very first responses (fill-till-full), and refinement engages only *after* it's full, so there's no warm-up or learning penalty before the table is useful.
- **High compression with small state.** Gain-weighted swapping lets a small table reach what a much larger fill-only table needs — 4 KB refined ≈ 16 KB fill-only below. And the shadow cache is encoder-only, so the *decoder* footprint stays at the negotiated table size; the cost of refining is paid solely by the sender that opts into it.
- **Adapts on long-running connections.** A full table is no longer frozen — it follows the evolving hot-header set, which is what matters for long-lived API/RPC connections; the grow-and-freeze keeps that adaptation bounded in cost.
- **Low head-of-line-blocking risk.** The hysteresis (2× margin plus a repeat-evidence gate) keeps the encoder from thrashing the table, so encoder-stream traffic stays a small fraction of the header bytes it saves at the relevant table sizes. Sparse, stable inserts mean fewer header blocks reference a not-yet-received entry, limiting the QPACK head-of-line blocking that aggressive insertion would otherwise cause.

### Results — `fb-resp-hq`, immediate-ack (text baseline 340,737 B)

| table | refine | header B | encoder B | total B | total/text |
|------:|:------:|--------:|----------:|--------:|-----------:|
| 0 (static) | — | 207,109 | 0 | 207,109 | 60.8% |
| 4 KB | off | 70,320 | 1,971 | 72,291 | 21.2% |
| 4 KB | **on** | 47,970 | 3,796 | 51,766 | **15.2%** |
| 8 KB | off | 57,922 | 3,951 | 61,873 | 18.2% |
| 8 KB | **on** | 41,209 | 5,669 | 46,878 | **13.8%** |
| 16 KB | off | 44,393 | 7,779 | 52,172 | 15.3% |
| 16 KB | **on** | 35,673 | 8,909 | 44,582 | **13.1%** |
| 64 KB | off | 30,652 | 17,080 | 47,732 | 14.0% |
| 64 KB | **on** | 27,779 | 17,169 | 44,948 | 13.2% |

- Refinement lowers total bytes at every size: **−28% (4 KB) → −6% (64 KB)** — a small-table win.
- **4 KB refined ≈ 16 KB fill-only** (51.8 KB vs 52.2 KB total) at ~⅓ the table memory.
- Best total is **16 KB refined (13.1%)**; bigger tables stop helping once the one-time encoder-stream cost grows (fill-only 64 KB is *worse* than refined 8 KB).

### Caveat
These numbers use one trace with **immediate acks** — the friendliest schedule for eviction. Sustained in-flight pinning (which the offline simulator also didn't model) could reduce the gain; that's the next thing to measure with a delayed-ack model.

### Config
`refine_after_full` in `h2o_http3_qpack_context_t`, default on. Server: `qpack-encoder-refine: OFF|ON` (applies to responses). httpclient: `--http3-qpack-encoder-refine <0|1>` (applies to requests).


kazuho:
## Comparison vs HPACK (h2o's HTTP/2 encoder) — corrected

*This replaces an earlier version of this comment that credited QPACK's static table for ~half the win. That was wrong; the corrected analysis is below.*

Added `--hpack` to `t-qif` (encoder-only, responses) so the same trace flattens through h2o's HPACK encoder. Same `fb-resp-hq` trace, same text baseline (340,737 B). HPACK has no encoder stream — inserts are inline — so its number is the full wire cost, directly comparable to QPACK *total*.

| config | header B | encoder B | total | /text |
|---|--:|--:|--:|--:|
| **HPACK** (h2o, 32-entry table) | 90,448 | 0 | 90,448 | 26.5% |
| QPACK 4K, fill-only | 70,320 | 1,971 | 72,291 | 21.2% |
| QPACK 4K, refine | 47,970 | 3,796 | 51,766 | 15.2% |
| QPACK 16K, refine | 35,673 | 8,909 | 44,582 | 13.1% |

### Same entry count — is QPACK just holding more?

The 4K QPACK table holds ~50–60 entries vs HPACK's 32, so the obvious question is whether QPACK simply has more room. Capping QPACK at **32 entries** too (matched to HPACK) answers it:

| 32 entries | header B | encoder B | total | /text |
|---|--:|--:|--:|--:|
| HPACK | 90,448 | 0 | 90,448 | 26.5% |
| QPACK, fill-only | 71,500 | 1,650 | 73,150 | 21.5% |
| QPACK, refine | 50,330 | 2,765 | 53,095 | 15.6% |

At the **same 32 entries**, fill-only is −19% and refine is −41% vs HPACK — so the win is policy, not capacity. And refine-32 (53,095) ≈ refine-4K (51,766, within 2.5%): refinement extracts almost all of its value from 32 well-chosen entries, i.e. it keeps the *right* entries rather than merely more of them.

### Where the gap actually comes from

h2o's HPACK encoder caps its dynamic table at **32 entries** (it linear-searches the table per field in `do_encode_header`, so it can't afford more). That cap is essentially the whole story. Sweeping the cap on the real HPACK encoder (byte capacity pinned at 64K so only entry count matters):

| HPACK entry cap | total | excess over floor | excess × N |
|---:|---:|---:|---:|
| 32 | 90,448 | 46,588 | 1,490,816 |
| 48 | 76,218 | 32,358 | 1,553,184 |
| 64 | 68,135 | 24,275 | 1,553,600 |
| 89 | 60,837 | 16,977 | 1,510,953 |
| 128 | 52,740 | 8,880 | 1,136,640 |
| 256 | 48,140 | 4,280 | 1,095,680 |
| ∞ | 43,860 | 0 | 0 |

Two facts fall out:

1. **HPACK's deficit is FIFO thrash that scales as 1/N.** `excess × N` is ~constant (~1.5M) while the table is smaller than the working set. Mechanism: HPACK evicts strictly in insertion order, and an Indexed Header Field reference does **not** refresh an entry (RFC 7541 §6.1: "does not change the dynamic table"). So a hot pair ages out every ~`N/r` responses no matter how often it is used, then has to be re-sent as a literal — reinsertion frequency ∝ `1/N`. (It decays faster than 1/N past ~128 entries because a larger table turns more literals into hits, lowering the insert rate `r` as well, and reaches 0 once the table holds the whole working set.)

2. **The floors converge.** Uncapped HPACK bottoms out at **43,860** — within 1.6% of QPACK 16K refine (44,582), and actually a hair *lower* (QPACK pays ~8.9K of encoder stream that HPACK folds inline). At equal table room HPACK's per-field coding is as good as QPACK's: the Huffman tables are identical, and the static-table difference is only ~**300 B one-time** (just 12 distinct (name,value) pairs in this trace live in QPACK's static table but not HPACK's). So the win is **not** "QPACK is a better coder," and it is **not** the static table.

### What QPACK actually buys — decomposed at equal entries

Capping QPACK to HPACK's 32 entries isolates *why* it wins, independent of table size. Decomposing the 90,496 → 53,095 gap, all at 32 entries:

| stage | total | Δ | needs Duplicate? |
|---|--:|--:|:--:|
| h2o HPACK (indexes every literal, evicts forever) | 90,496 | — | |
| **+ freeze when full** (switch to literal-without-indexing) | 75,440 | **−15,056** | **no** |
| QPACK fill-32 (freeze + richer static table + enc-stream) | 73,150 | −2,290 | no |
| QPACK refine-32 (admission + adaptive swap) | 53,095 | −20,055 | swap: yes |

Three levers, in order of size:

1. **Freeze (~40%, Duplicate-free).** The single biggest cause of h2o HPACK's deficit is that it never stops indexing — it emits Literal-with-Incremental-Indexing (`0x40`) for every compressible value and therefore evicts on every insert, forever. Switching to Literal-without-Indexing (`0x00`) once the table is full — i.e. QPACK's fill-till-full *freeze* — recovers ~15K **with no protocol feature, no admission logic, and no Duplicate.** HPACK-the-protocol already permits this; h2o's encoder simply doesn't do it.
2. **Admission + adaptive swap (~54%).** Shadow-rank admission keeps one-shot/rare values *out* of the table (they are the churn source), and the hysteretic swap upgrades the resident set as repeat-evidence accrues. The admission half is also Duplicate-free; only the *adaptive relocation* needs the protocol instruction below.
3. **Static table + encoding (~6%).** QPACK's 99-entry static table and post-base indexing, ~2K here.

**The instruction the swap needs.** To drop a low-value entry buried behind still-useful ones (eviction is FIFO from the front in QPACK too), the encoder emits **Duplicate** to relocate the keepers to the tail, then lets the front prefix evict — a reinsert that costs ~1–2 bytes on the encoder stream and does **not** re-send the value. HPACK's only "reinsert" is Literal-with-Incremental-Indexing, which re-sends the whole value. So Duplicate is what lets a *full* table keep tracking a *drifting* working set cheaply — it is the enabler for staying optimal over a long connection, not the main reason QPACK beats h2o's HPACK on this (stationary) trace. That is mostly the freeze.

Takeaways by reader: for an HPACK encoder author, the cheap ~15K win is **freeze when full** (no spec change). For a protocol comparison, QPACK's **Duplicate** is the affordance that makes adaptive, churn-free table maintenance possible at all.

*(All numbers immediate-ack — the friendly case for eviction; same caveat as the PR description. `--hpack` mode and the trace are in the branch; the entry-cap sweeps and the HPACK freeze test used throwaway `getenv` overrides that are not committed. The shipped default is now an 8KB encoder table with a `max(32, capacity/128)` entry limit → 64 entries.)*

kazuho:
Note on the results table in the description: those numbers predate the `max(32, capacity/128)` encoder entry cap that this branch now includes, so they no longer match `t-qif`. The cap limits how many entries a small table holds (4 KB → 32, 8 KB → 64, 16 KB → 128), which reduces inserts — fewer encoder-stream bytes, slightly more header bytes. The `static` and `64 KB` rows are unaffected (no table / the 512-entry cap never binds) and still match.

Current `t-qif` (`fb-resp-hq`, immediate-ack, text baseline 340,737 B):

| table | refine | header B | encoder B | total B | total/text |
|------:|:------:|--------:|----------:|--------:|-----------:|
| 0 (static) | — | 207,109 | 0 | 207,109 | 60.8% |
| 4 KB | off | 71,500 | 1,650 | 73,150 | 21.5% |
| 4 KB | **on** | 50,330 | 2,765 | 53,095 | **15.6%** |
| 8 KB | off | 57,424 | 3,350 | 60,774 | 17.8% |
| 8 KB | **on** | 41,668 | 5,137 | 46,805 | **13.7%** |
| 16 KB | off | 49,970 | 5,602 | 55,572 | 16.3% |
| 16 KB | **on** | 39,755 | 6,818 | 46,573 | **13.7%** |
| 64 KB | off | 30,652 | 17,080 | 47,732 | 14.0% |
| 64 KB | **on** | 27,779 | 17,169 | 44,948 | **13.2%** |

Refinement still lowers total bytes at every size, and 4 KB refined (53,095) still beats 16 KB fill-only (55,572) at ~⅓ the table memory. The one conclusion that shifts under the entry cap: the best total is now **64 KB refined (44,948, 13.2%)** rather than 16 KB; 16 KB and 8 KB refined land about even (~13.7%).

PR #3611, opened 2026-06-23
main.c: Support listening on MPTCP socket

ddiproietto:


kazuho:
Thank you for the PR.

I've changed so that Multipath TCP is enabled only when explicitly configured as such; i.e., `-DH2O_WITH_MPTCP=1`. The rationale is that it is currently only supported by `--mode worker`.

To support Multipath TCP in daemon and master modes, we need to make changes to server-starter. If necessary, someone can do that as a follow-up.

ddiproietto:
Makes sense, thanks for the review and the changes!

PR #3529, opened 2025-10-21
Implement on-the-fly zstd compression

Ramblurr:
This PR adds on-the-fly zstd compression (fixes #3386)

The PR is split into three commits that could be squashed if needed:

1. the code changes to add the feature itself
    - The tests t/50compress-hint.t and t/50compress.t have been updated also
2. updating documentation
3. important the upstream zstd repo into deps/zstd

I've followed the CONTRIBUTING as much as possible, but this is my first contribution to libh2o, so please let me know if I did anything incorrectly or if I could improve something.

PR #3612, opened 2026-06-24
Use brotli library from the system

deweerdt:


Ramblurr:
@deweerdt @kazuho I've had a PR adding zstd open for some time (https://github.com/h2o/h2o/pull/3529), would you consider a revised PR that uses the zstd system lib like this one does for brotli?

deweerdt:
@Ramblurr that would make sense to me, thanks for offering! What do you think @kazuho ?

